PR #27, opened 2024-04-25
Deflate large .nc files with ncks -4 -L 4

andrew-c-ross:
To help speed up checkouts and keep this repository from getting too large, this PR deflates the larger netCDF files in the repository with `ncks -4 -L 4`. According to the GitHub changed files this saves 85 MB. Deflation is lossless so this shouldn't change any of the diagnostic results.


andrew-c-ross:
I'm not sure about the nuances of Git though. Will these large files still lurk in the repository and get downloaded with new clones? But maybe with a partial or shallow clone it would help the automated tests at least.

yichengt900:
@andrew-c-ross, thanks for the changes. I think the checkout process in our GitHub Actions seems to be okay (~16s-19s) at this moment, but it makes sense to deflate the larger NetCDF files. Right now, the major bottleneck is the build process, especially when we try to build MOM6SIS2 with production flags (debug mode is okay). This is on my TODO list but I have not had the chance to do it yet.

yichengt900:
Also this is good to remind me I should remove  `get_glorys_data_ci ` test from each PR. We only need to run this test daily.

yichengt900 (approved):

PR #35, opened 2024-05-03
Add dumbbell test case

yichengt900:
As titled, this PR adds a simple `dumbbell ` z* coordinate test case and associated testing workflow. It is inexpensive and can be a useful debugging test case for OBC-related issues. We also updated our MOM6 and ocean_BGC to the latest tags.


theresa-morrison:
This looks good! I was able to run both the full domain and sub region test cases. 

I have a couple suggestions:
(1) do we need to use a layout option or can we leave that unspecified? I don't know how this impacts the CI, but I find not having to change the layout file makes manually checking the reproducibility across PE counts easier. 
(2) additionally/related, is this case meant to be an example for users or just for internal testing? Either way, I think it might be helpful to have a short read me in the dumbbell directory that lists the steps to run this experiment (essentially what is in the driver.sh script). I can add that to this PR or as a new PR once this one is accepted.

yichengt900:
Thx @theresa-morrison .

1. Good point. I can simply remove MOM_layout from `input.nml` so we can randomly pick desired PEs during runtime.
2. Sounds good to me. I think it might be easier to merge this PR first and then you can create another PR for more descriptions.

theresa-morrison (approved):
looks good!

PR #50, opened 2024-06-12
Add empty figures directories

andrew-c-ross:
The diagnostic scripts assume the existence of directories named 'figures' where they can save the figures to. This PR just adds two empty directories so that users don't need to make them before running the diagnostics.


yichengt900 (approved):
@andrew-c-ross , good catch. Perhaps we could consider to have a simple bash script as an example to call all these  scripts. Approve for now.

PR #52, opened 2024-06-13
Increased job runtimes to 13 hours 

HarrisLongfield:
Changed job time from 8:00:00 to 13:00:00 to prevent a jobtime failure while running the model. It seems like the model takes longer to run than it used to -- ~ 1 hour per month.


yichengt900:
Hi @HarrisLongfield, thanks for opening this PR. Good catch. The main reason for this change is that we have reduced the model time step from `DT=600s` to `DT=400s`.

yichengt900 (approved):

PR #60, opened 2024-07-09
Adds FRE canopy YAMLs for building the model

thomas-robinson:
Adds yaml files that can be used for building the executable with FRE Canopy `fre make`.   This can be used to make a container for the model.

It's missing directions and several other things that are probably important.


yichengt900:
Since #131 has been merged, I will close this draft PR for now. @thomas-robinson, thank you again for sharing an excellent example of how to use the new FRE and how the container works.

PR #62, opened 2024-07-11
Add environment and makefile for GNU 12 on GFDL workstations

andrew-c-ross:
As titled, this PR adds files for using the GNU 12 compilers on GFDL workstations. The other existing environments for gfdl-ws don't seem to work.


yichengt900:
@andrew-c-ross,

Thank you for adding the environment and makefiles for GNU12 on GFDL WS. I tested them with the 1D case, and everything works fine. Out of curiosity, have you also tried the Intel 2022 compiler on GFDL WS?

yichengt900 (approved):

andrew-c-ross:
I don't think I've tried Intel 2022. Based on the modules that are available and what they were compiled with I thought GNU 12 seemed like the one most likely to work.

PR #65, opened 2024-07-19
Add generic age tracer in dumbbell case

yichengt900:
As titled. Now We have included a generic age tracer in the dumbbell test case. This addition is a preparatory step for future debugging and development of the generic tracer OBC.


theresa-morrison:
I tested these changes and the generic tracer seems to work!

theresa-morrison (approved):

yichengt900:
Thanks @theresa-morrison. Will merge this PR now.

PR #74, opened 2024-08-05
Add config file for diagnostic scripts and use it to generalize sst_eval.py

uwagura:
This pull request adds a config.yaml file to the physics diagnostics directory to make it easier to run the diagnostics scripts across domains by removing hardcoded variables. As a first test for the new config file, this PR also modifies the sst_eval.py script to use variables supplied by the config file. The script produces the same output as before for the NWA domain, but it can now be used across other domains as well by modifying the variables within the config file. Below is are the outputs from the script in the NWA and NEP domains using the new config file 

sst_eval
sst_eval_nep_align
sst_eval_better_lims
sst_eval_arctic

### DIFF
--- a/diagnostics/physics/sst_eval.py
+++ b/diagnostics/physics/sst_eval.py
@@ -20,7 +20,7 @@ from plot_common import annotate_skill, autoextend_colorbar, corners, get_map_no
 def plot_sst_eval(pp_root, config):
     model = open_var(pp_root, config['domain'], 'tos')
     model_grid = xarray.open_dataset( config['model_grid'] )
-    target_grid = model_grid[config['rename_vars']].rename(config['rename_map'])
+    target_grid = model_grid[config['rename_map'].keys()].rename(config['rename_map'])
     model_ave = model.mean('time').load()
 
     glorys = xarray.open_dataset( config['glorys'] ).squeeze()['thetao'] #.rename({'longitude': 'lon', 'latitude': 'lat'})
@@ -48,7 +48,7 @@ def plot_sst_eval(pp_root, config):
     oisst_ave = oisst.mean('time').load()
     mom_rg = mom_to_oisst(model_ave)
     delta_oisst = mom_rg - oisst_ave
-    fig = plt.figure(figsize=(11, 14))
+    fig = plt.figure(figsize=(config['fig_width'], config['fig_height']))
     grid = AxesGrid(fig, 111,
         axes_class=(GeoAxes, dict(projection=ccrs.PlateCarree())),
         nrows_ncols=(2, 3),
@@ -88,7 +88,7 @@ def plot_sst_eval(pp_root, config):
     # Model - OISST
     grid[2].pcolormesh(oisst_lonc, oisst_latc, delta_oisst, **bias_common)
     grid[2].set_title('(c) Model - OISST')
-    annotate_skill(mom_rg, oisst_ave, grid[2], dim=['lat', 'lon'])
+    annotate_skill(mom_rg, oisst_ave, grid[2], dim=['lat', 'lon'], x0=config['text_x'], y0=config['text_y'], yint=config['text_yint'])
 
     # GLORYS
     p1 = grid[4].pcolormesh(glorys_lonc, glorys_latc, glorys_ave, cmap=cmap, norm=norm)
@@ -102,7 +102,7 @@ def plot_sst_eval(pp_root, config):
     cbar2.ax.set_xlabel('SST difference (°C)')
     cbar2.ax.set_xticks([-2, -1, 0, 1, 2])
     grid[5].set_title('(e) Model - GLORYS12')
-    annotate_skill(model_ave, glorys_rg, grid[5], weights=model_grid.areacello)
+    annotate_skill(model_ave, glorys_rg, grid[5], weights=model_grid.areacello, x0=config['text_x'], y0=config['text_y'], yint=config['text_yint'])
 
     for ax in grid:
         ax.set_xlim(config['x']['min'], config['x']['max'])
--- a/diagnostics/physics/sst_eval.py
+++ b/diagnostics/physics/sst_eval.py
@@ -12,10 +12,13 @@ import matplotlib.pyplot as plt
 from mpl_toolkits.axes_grid1 import AxesGrid
 import numpy as np
 import xarray
-import xesmf
+import logging
 
-from plot_common import annotate_skill, autoextend_colorbar, corners, get_map_norm, open_var
+from plot_common import annotate_skill, autoextend_colorbar, corners, get_map_norm, open_var, load_config, process_oisst, process_glorys
 
+# Configure logging for sst_eval
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


yichengt900:
Hi @gabyneg and @theresa-morrison,

@uwagura is currently working on generalizing our diagnostic scripts to support multiple regional domains. Could you please share the post-processing (PP) paths on archive for the Hawaii and Arctic domains? They don't need to be the best results; we just want to test the script and ensure it functions correctly across different domains. Thanks!

gabyneg:
Hi @yichengt900 and @uwagura, my PP step has not been functioning properly, so I have been using my output from my history files e.g. : "/archive/g1n/fre/cefi/MHI/CEFI_MHI_COBALT_V2/gfdl.ncrc5-intel22-prod/history"

yichengt900:
@gabyneg, thanks, and no worries. If you are willing to share your example XML, perhaps @uwagura can try to help implement the PP step for the Pacific Islands.

uwagura:
I can definitely look into the pp issue @gabyneg . And thanks for sharing the path to the history files!

gabyneg:
@uwagura Thank you! Here are my xml files (physics only) "/gpfs/f5/cefi/proj-shared/MHI/xmls/MHI/MHI_physics.xml", &  (physics + bgc): "/gpfs/f5/cefi/proj-shared/MHI/xmls/MHI/CEFI_MHI_cobalt.xml".

theresa-morrison:
@yichengt900 Here is a path with pp for the full simulation period:
/archive/Theresa.Morrison/fre/cefi/ARC12/ARC12_51/gfdl.ncrc5-intel22-prod/pp

Note that Kate has adjusted a lot of parameters since I ran this experiment, so it is not the up-to-date state of the Arctic configuration!

diagnostics/physics/sst_eval.py
```diff
@@ -17,23 +17,25 @@
...
 
 
-def plot_sst_eval(pp_root):
-    model = open_var(pp_root, 'ocean_monthly', 'tos')
-    model_grid = xarray.open_dataset('../data/geography/ocean_static.nc')
-    target_grid = model_grid[['geolon', 'geolat', 'geolon_c', 'geolat_c']].rename({'geolon': 'lon', 'geolat': 'lat', 'geolon_c': 'lon_b', 'geolat_c': 'lat_b'})
+def plot_sst_eval(pp_root, config):
+    model = open_var(pp_root, config['domain'], 'tos')
+    model_grid = xarray.open_dataset( config['model_grid'] )
+    target_grid = model_grid[config['rename_vars']].rename(config['rename_map'])
```
andrew-c-ross, line 23: It should be possible to do something like `model_grid[config['rename_map'].keys()].rename(etc)` or `model_grid[list(config['rename_map'].keys())].rename(etc)` so that you don't need the extra `rename_vars` in the yaml

diagnostics/physics/sst_eval.py
```diff
@@ -101,24 +102,28 @@ def plot_sst_eval(pp_root):
...
 if __name__ == '__main__':
     from argparse import ArgumentParser
+    import yaml
     parser = ArgumentParser()
-    parser.add_argument('pp_root', help='Path to postprocessed data (up to but not including /pp/)')
+    parser.add_argument('-p','--pp_root', type=str, help='Path to postprocessed data (up to but not including /pp/)')
+    parser.add_argument('-c','--config', type=str, help='Path to config.yaml file containing relevant paths for diagnostic scripts')
     args = parser.parse_args()
-    plot_sst_eval(args.pp_root)
+    with open(args.config, 'r') as f:
```
yichengt900, line 127: @uwagura, I like the idea of using a YAML file to read the control parameters. I believe we can make this part as a function and move it to `plot_common.py`, so it can be reused by other scripts in the future:
Example in `plot_common.py`:
```
import logging
import yaml

# Configure logging for plot_common
logger = logging.getLogger(__name__)

def load_config(config_path: str):
    """Load the configuration file."""
    try:
        with open(config_path, 'r') as f:
            config = yaml.safe_load(f)
            logger.info(f"Loaded configuration from {config_path}")
        return config
    except Exception as e:
        logger.error(f"Error loading configuration from {config_path}: {e}")
        raise
```

diagnostics/physics/sst_eval.py
```diff
@@ -17,23 +17,25 @@
...
+    except:
+        glorys_lonc, glorys_latc = corners(glorys.longitude, glorys.latitude)
     glorys_ave = glorys.mean('time').load()
     glorys_to_mom = xesmf.Regridder(glorys_ave, target_grid, method='bilinear', unmapped_to_nan=True)
     glorys_rg = glorys_to_mom(glorys_ave)
     delta_glorys = model_ave - glorys_rg
-
     oisst = (
-        xarray.open_mfdataset([f'/work/acr/oisstv2/sst.month.mean.{y}.nc' for y in range(1993, 2020)])
+        xarray.open_mfdataset([config['oisst']+f'sst.month.mean.{y}.nc' for y in range(1993, 2020)])
```
yichengt900, line 36: @uwagura, do you think it would be a good idea to move the OISST and GLORYS data processing into separate functions? It might make it easier to maintain and keep things organized. We can try something like the following:

```
def process_oisst(config, target_grid, model_ave):
    """Process OISST data."""
    try:
        oisst = (
            xarray.open_mfdataset([config['oisst'] + f'sst.month.mean.{y}.nc' for y in range(1993, 2020)])
            .sst
            .sel(lat=slice(config['lat']['south'], config['lat']['north']), lon=slice(config['lon']['west'], config['lon']['east']))
        )
        oisst_lonc, oisst_latc = corners(oisst.lon, oisst.lat)
        oisst_lonc -= 360
        mom_to_oisst = xesmf.Regridder(
            target_grid,
            {'lat': oisst.lat, 'lon': oisst.lon, 'lat_b': oisst_latc, 'lon_b': oisst_lonc},
            method='conservative_normed',
            unmapped_to_nan=True
        )
        oisst_ave = oisst.mean('time').load()
        mom_rg = mom_to_oisst(model_ave)
        logger.info("OISST data processed successfully.")
        return oisst_ave, mom_rg, oisst_lonc, oisst_latc
    except Exception as e:
        logger.error(f"Error processing OISST data: {e}")
        raise
```

diagnostics/physics/sst_eval.py
```diff
@@ -70,55 +75,70 @@ def plot_sst_eval(pp_root):
...
+
+    plt.savefig(config['figures_dir']+'sst_eval.png', dpi=300, bbox_inches='tight')
+    logger.info("Successfully saved figure")
 
 
 if __name__ == '__main__':
     from argparse import ArgumentParser
     parser = ArgumentParser()
-    parser.add_argument('pp_root', help='Path to postprocessed data (up to but not including /pp/)')
+    parser.add_argument('-p','--pp_root', type=str, help='Path to postprocessed data (up to but not including /pp/)')
```
andrew-c-ross, line 140: Since these two command line arguments are flags but are not optional, you could add `required=True` to the `add_argument` call so that it will generate an exception right away if both command line args aren't included.

andrew-c-ross (approved):

yichengt900 (approved):
@uwagura , thanks for your great work! The new plots look good. Approve.

PR #78, opened 2024-08-13
Preparation for Transition to C6 (Regression testing first, WIP)

yichengt900:
As titled.

### DIFF
--- a/builds/gaea/ncrc6.intel23.env
+++ b/builds/gaea/ncrc6.intel23.env
@@ -1,0 +1,11 @@
+source $MODULESHOME/init/tcsh
+module unload cray-netcdf cray-hdf5 fre
+module unload PrgEnv-pgi PrgEnv-intel PrgEnv-gnu PrgEnv-cray
+module load PrgEnv-intel/8.5.0
+module unload intel intel-classic intel-oneapi
+module load intel-classic/2023.2.0
+module load cray-hdf5/1.12.2.11
+module load cray-netcdf/4.9.0.9
+module load libyaml/0.2.5
+module list
+ftn --version
--- a/builds/gaea/ncrc6.intel23.mk
+++ b/builds/gaea/ncrc6.intel23.mk
@@ -1,0 +1,279 @@
+# Template for the Intel Compilers on a Cray System
+#
+# Typical use with mkmf
+# mkmf -t ncrc-cray.mk -c"-Duse_libMPI -Duse_netCDF" path_names /usr/local/include
+
+############
+# Commands Macros
+############
+FC = ftn
+CC = cc
+LD = ftn $(MAIN_PROGRAM)
+
+#######################
+# Build target macros
+#
+# Macros that modify compiler flags used in the build.  Target
+# macrose are usually set on the call to make:
+#
+#    make REPRO=on NETCDF=3
+#
+# Most target macros are activated when their value is non-blank.
+# Some have a single value that is checked.  Others will use the
+# value of the macro in the compile command.
+
+DEBUG =              # If non-blank, perform a debug build (Cannot be
+                     # mixed with REPRO or TEST)
+
+REPRO =              # If non-blank, erform a build that guarentees
+                     # reprodicuibilty from run to run.  Cannot be used
+                     # with DEBUG or TEST
+
+TEST  =              # If non-blank, use the compiler options defined in
+                     # the FFLAGS_TEST and CFLAGS_TEST macros.  Cannot be
+                     # use with REPRO or DEBUG
+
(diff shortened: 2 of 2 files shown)


xmls/xml_include/xml_building_blocks/platforms.xml
```diff
@@ -30,6 +30,37 @@
...
+    </directory>
+    <property name="EXEC_ROOT"         value="/gpfs/f6/$(project)/scratch/$USER/$(stem)/$(BUILD_DATE)"/>
+    <property name="F2003_FLAGS"       value=" -DINTERNAL_FILE_NML -g "/>
+    <property name="AWG_INPUT_HOME"    value="$(AWG_INPUT_HOME_GAEA)"/>
+    <property name="NB_ROOT"           value="/nbhome/$USER/$(FRE_STEM)$(PLATINFO)/$(name)"/>
+    <property name="ARCHIVE_NNZ"        value=""/>
+    <property name="REFERENCE"         value=""/>
+    <property name="FMS_ARCHIVE_ROOT"    value=""/>
+    <property name="CMIP6_ARCHIVE_ROOT"  value=""/>
+    <property name="NCORE_PER_NODE"      value="128"/>
```
andrew-c-ross, line 55: I don't know if this even gets used anywhere, but technically, is c6 192 cores per node?

xmls/xml_include/xml_building_blocks/platforms.xml
```diff
@@ -30,6 +30,37 @@
...
+    <property name="NPES_IO_LAND"         value="18"/>
+    <csh><![CDATA[
+      module load git
+      module unload darshan
```
andrew-c-ross, line 59: darshan is now darshan-runtime (it has been on c5 too, not sure when it was renamed)

xmls/xml_include/xml_building_blocks/platforms.xml
yichengt900: @andrew-c-ross, that’s correct. I thought I had changed it, but apparently, I missed it. Good catch.

xmls/xml_include/xml_building_blocks/platforms.xml
yichengt900: Yup another good catch. Actually I don't think we really need this line since it is not shown in MDT's platforms.xml for c5. I will just remove it for now.

andrew-c-ross: There was a period in time where darshan would randomly cause crashes, and unloading it was the unofficial solution. I'm not sure if this is still the case, but I've always continued to unload it out of superstition

yichengt900: Ok no problem then I will add it back with correct naming.

xmls/NWA12/CEFI_NWA12_cobalt.xml
```diff
@@ -94,7 +95,7 @@ frecheck -r layout -p ncrc5.intel22 -x CEFI_NWA12_cobalt.xml CEFI_NWA12_COBALT_V
   <property name="STATIC" value=""/>
 
   <!-- Path to world-shared location with common NWA12 input files -->
-  <property name="NWA12_WORLD" value="/gpfs/f5/gfdl_med/world-shared/northwest_atlantic/nwa12_input"/>
+  <property name="NWA12_WORLD" value="/gpfs/f6/ira-cefi/world-shared/northwest_atlantic/nwa12_input"/>
```
andrew-c-ross, line 98: What do you think about putting properties like this with the platforms.xml, so that it can point to f5 for c5 platforms and f6 for c6? 

That or where we have `<dataSource>`, I think something like `<dataSource platform='ncrc6.intel23'>` would work

yichengt900: Thanks for the suggestion, @andrew-c-ross. I moved the data folder properties to the platforms.xml now.

andrew-c-ross (approved):
Everything here seems to work for me. Thanks Yi-Cheng! The only thing I haven't tested out yet is post-processing.

Side note, that is the best Gulf Stream and cold pool simulation I've ever seen with this model!

charliestock:
Hi Yi-Cheng - thanks for all of your hard work on this.  Amazing progress, and I have to agree with Andrew.  That Gulf Stream looks amazing!  Is that result deterministic or just a lucky spin?

yichengt900:
@charliestock I would say it's half-half. The Glorys data nudging likely played a role, but as I recall, we didn't achieve such good results even with nudging in the past?

PR #85, opened 2024-08-22
Turn on PHA_MLD_CALC and associated runtime parameters to allow MOM6 to calculate a photoacclimation MLD

yichengt900:
As titled. this PR updates the model configurations to enable MOM6 to calculate the photoacclimation mixed layer depth (MLD). We have also updated the `diag_table` to illustrate how to output both the default `MLD003` and a user-specified MLD.

While we have passed `photo_acc_dpth` to `generic_tracer_source`, currently we do not pass `photo_acc_dpth`  further into `generic_COBALT_update_from_source` for the final photoacclimation  calculation:
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/dc276573f5cca0a63312ba97d0c1b7f6c5c81ac0/generic_tracers/generic_tracer.F90#L554-L558
 so the answers remain the same in this PR. @theresa-morrison has another [PR](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/pull/91#event-13983248127) now  which will eventually bring in MOM6-calculated photoacclimation MLD into COBALT and we except the answer changes for that PR.


xmls/NWA12/diag_tables/diag_table_MOM6_2023-03
theresa-morrison, line 1: Should we consider renaming this file to preserve the version that is intended to be run with older versions of MOM6 without the depth correction for the MLD?

theresa-morrison (commented):
The changes for the PHA_MLD_CALC and the MLD diagnostic look correct. 

I would recommend updating the name of the diag_table file because it has different entries than the 2023-03 version. If there is a reason not to rename it to a more recent date I would suggest we remove "2023-03" from the name at least.

Note that with these changes MLD_DBM is what should be compared to the de Boyer Montegut dataset. The MLD_003 diagnostic should be consistent with previous results as long as no answer-changing flags are turned on.

xmls/NWA12/diag_tables/diag_table_MOM6_2023-03
yichengt900: Good point. I have created a new diag_table and preserved the old one. I also noticed that this may lead to inconsistencies in the NWA XML, as the PPAN configuration still references @andrew-c-ross's diag_table. While this should not significantly impact the workflow, I will sync with @andrew-c-ross later to ensure consistency.

theresa-morrison (approved):

PR #88, opened 2024-09-10
Temporal solution for data transfer on C6

yichengt900:
Following the recent maintenance, the `dtn_f5_f6` partition appears to be malfunctioning on C6, which has caused data transfer issue for users attempting to load `fre/test` and run the FRE workflow on C6. To mitigate this, I’ve created a temporary version of `fre/test` with a hotfix (use `rdtn_c5` and `rdtn_c5` instead) in my home directory to ensure that we can continue our work on C6 without disruption.

FYI @andrew-c-ross , @charliestock , @gabyneg , @theresa-morrison , @uwagura

### DIFF
--- a/xmls/NWA12/CEFI_NWA12_cobalt.xml
+++ b/xmls/NWA12/CEFI_NWA12_cobalt.xml
@@ -9,7 +9,7 @@ On GFDL Analysis:
 module load fre/test
 
 On Gaea:
-module use -a /ncrc/home2/fms/local/modulefiles
+module use -a /ncrc/home2/Yi-cheng.Teng/local/modulefiles
 module load fre/test
 
 Compile and build mom6-sis2-cobalt:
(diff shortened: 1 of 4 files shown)


andrew-c-ross:
Thanks Yi-Cheng! Have you run into any issues using bronx-22 on c6 instead of test? Chris Blanton said it should be usable now. I tried reverting to it, but I'm not sure if my output is transferring.

yichengt900:
> Thanks Yi-Cheng! Have you run into any issues using bronx-22 on c6 instead of test? Chris Blanton said it should be usable now. I tried reverting to it, but I'm not sure if my output is transferring.

@andrew-c-ross, I also tested both bronx-21 and 22, and I've identified another issue, which you can find here: https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/regional-mom6-xml/pull/29. For some reason, neither bronx-21 nor bronx-22 is handling the restart archive correctly (both ASCII and history files are being archived and transferred if you specified --transfer argument). At the moment, my `fre/test` is the only version that works properly. I plan to open a helpdesk ticket later to address the issue with bronx-21 and bronx-22.

andrew-c-ross:
Ahh, thanks. I'll give your fre/test a try.

PR #90, opened 2024-09-12
Add basic tide only example for NWA12

andrew-c-ross:
This PR adds a single layer, tides-only configuration for NWA12. The settings are based on the [MOM6_examples tides_025 case](https://github.com/NOAA-GFDL/MOM6-examples/tree/68773d337401239f010a4a08747b0ea0c8647022/ocean_only/tides_025), with the addition of open boundary conditions and a few minor changes, while the grid is from NWA12. All of the netcdf files needed for this setup are already provided in the datasets tar file, so this is a lightweight addition. Among other things, this configuration could be useful for testing out tidal boundary conditions and trying to debug any tide issues in regional models. It should be compiled in ocean-only mode, so it's a little faster and simpler. I'm specifically sending this to help test out [this PR](https://github.com/c2xu/MOM6/tree/c2xu/nodal_modulation), though.


yichengt900:
Thanks, @andrew-c-ross! It's great to have an additional test for the tides-only test case. I can run this case on C6. Just wondering, do you think it’s a good idea to add a run script like `driver.sh` so users have a clear idea of how to run this test case?
`driver.sh`:
```
#!/bin/bash
#SBATCH --nodes=5
#SBATCH --time=60
#SBATCH --job-name="NWA12.tidesonly"
#SBATCH --output=NWA12.tidesonly_o.%j
#SBATCH --error=NWA12.tidesonly_e.%j
#SBATCH --qos=debug
#SBATCH --partition=batch
#SBATCH --clusters=c6
#SBATCH --account=ira-cefi

#
ntasks1=900

#
echo "Test started:  " `date`

#
echo "link datasets ..."
pushd ../
ln -fs /gpfs/f6/ira-cefi/world-shared/datasets ./
popd

#
echo "run tidesonly test ..."
srun --ntasks ${ntasks1} --cpus-per-task=1 --export=ALL ../../builds/build/gaea-ncrc6.intel23/ocean_only/debug/MOM6

echo "Test ended:  " `date`
```

andrew-c-ross:
Sure, I've added a script like that (I haven't tested it, though). This setup is actually lightweight enough that I've been running it on my workstation.

yichengt900:
> Sure, I've added a script like that (I haven't tested it, though). This setup is actually lightweight enough that I've been running it on my workstation.

Thanks, @andrew-c-ross , it looks good to me. Thanks again for this effort!

yichengt900 (approved):

Issue #87, opened 2024-09-06
Diagnostic scripts producing inconsistent output


uwagura:
Several scripts within the `diagnostics/physics` subfolder have been producing output that is inconsistent between runs. When I originally ran `sss_eval.py` on 07/29/2024, I produced the following figure:

![sss_eval](https://github.com/user-attachments/assets/c4755df0-7437-4a49-99c4-7bf2a0a3d064)

However, all recent runs (i.e from at least 08/29/2024 to present) produce the following output: 

![sss_eval_config](https://github.com/user-attachments/assets/6cd99171-d3b0-4ff8-9ba8-25166a120b38)

Note that the skill metrics are slightly different, and that the `model - oisst`,`model -glorys`, and `model` plots are all slightly different between these two runs.

Andrew previously ran the same script, which was published [here](https://gmd.copernicus.org/articles/16/6943/2023/gmd-16-6943-2023-discussion.html), and produced a figure with a slightly different `glorys` plot and slightly different skill metrics for 
`model-glorys` compared to both of the above runs

![figure04](https://github.com/user-attachments/assets/f25dc0e3-e038-4618-bb34-c2bffa9998c2)

A similar issue occurs with the `sst_eval` script. A run that took place around 07/29/2024 produced the following output: 

![sst_eval](https://github.com/user-attachments/assets/2e14621d-b6d9-4bed-99f5-22b4922d205c)

while a run from a few days later ( 08/01/2024 ) produced the following figure: 
![sst_eval](https://github.com/user-attachments/assets/69530d04-2a4b-43db-9ba7-893bc614b2c5)

The august runs of these plots were likely produced using Yi-Cheng's `setup` conda environment, which hasn't been modified recently. The july runs *most* likely used the same environment as well. Likewise, all of these runs used the same history files (currently listed at the top of each of these scripts), and these files have not been altered since the initial run. 

It's unclear what's causing the discrepency. MSD thinks the analysis nodes having slightly different firmware ( some nodes use [avx](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions) , while others don't ) may have contributed to this issue, or perhaps some other subtle hardware difference that python doesn't like. In my testing, I did not notice any difference between the nodes with different versions of `avx` ( listed [here](https://wiki.gfdl.noaa.gov/index.php/Post-processing_and_Analysis_User_Guide#cite_note-arch-4)). 

[...]

PR #91, opened 2024-09-12
Add generalized sss_eval script

uwagura:
This pull request adds config options to the config.yaml file to run `sss_eval.py` across domains. It also introduces helper functions to open and regrid regional climatology files from the National Centers for Environmental Information (NCEI), and modifies the previously introduced `process_glorys` and `process_oisst` scripts to select variables other than `thetao` from the glorys and oisst datasets. 

The new generalized version can reproduce it's previous output from the NWA domain ( i.e the "new" output mentioned in issue #87 ), and also works in the NEP domain. Since the NCEI does not provide regional climatologies for the Hawaiian Islands and does not provide decadal data for the Arctic, the new script has not been tested in those domains. 
sss_combine_coords
sss_eval_nep

A note about the NWA output: the previous version of this script opened the regional climatology files for the Northwest Atlantic using the the `compat='override'` argument to `xarray.open_mfdataset`. 
https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/ad3b53b6e889dc1649216e5533f14dfa9af936ff/diagnostics/physics/sss_eval.py#L40
This argument caused xarray to set the  data from `2005-2017` to `NaN` when merging the datasets, meaning that we were effectively just plotting `1995-2004` previously. This newer version instead uses `xarray.open_mfdataset( [files], combine = 'nested', concat_dim = 'time')` which should produce the same output as calling `xarray.open_mfdataset` without any options without throwing away any data. These arguments have the added benefit of allowing the `combine_regional_climatology` function to work with files with identical dimensions ( such as the regional climatologies for the NEP domain) , something which raises an error when using the xarray default argument `combine = 'combine_coords'`)

One last note about the NEP plot: the regional climatologies used in this plot cover slightly different periods since they are actually two domains stitched together ( a [Northern North Pacific](https://www.ncei.noaa.gov/access/nn-pacific-climate/bin/nnpregcl.pl?parameterOption=s) region covering 1995-2014, and a [Northeast Pacific](https://www.ncei.noaa.gov/access/ne-pacific-climate/bin/nepregcl.pl?parameter=s) region covering 1995-2012), meaning this plot is sort of comparing apples to oranges. The `combine_regional_climatology` function let's you do this since it calculates the period covered by each region from the file name so that it can apply the different weights to each region file. As with the last pull request, I wouldn't take the actual plot / skill metrics too seriously, but I would like to hear people's thoughts on whether or not `combine_regional_climatology` should allow you to input files covering different time periods.


andrew-c-ross:
Sorry for the bad news: I am getting sliiightly different results: an RMSE of 0.70 relative to the regional climatology instead of 0.71. The actual number is 0.7015044134823573, so it's not like it's close to rounding. 

sss_eval

andrew-c-ross:
Seems like the results are sensitive to what grid the regional climatologies are combined onto. I can reproduce the RMSE = 0.70 by changing the script I used in the paper to use the lat/lon values in this repo's config.yaml

diagnostics/physics/sss_eval.py
```diff
@@ -14,62 +14,54 @@
...
     # so mean ignores the nans from other regions and selects the available data.
-    regional_grid = {'lat': np.arange(10.05, 65.10, 0.1), 'lon': np.arange(-97.95, -39.90, 0.1)}
+    regional_grid = {'lat': np.arange( config['lat']['south'], config['lat']['north'], 0.1), 'lon': np.arange(config['lon']['west'], config['lon']['east'], 0.1)}
     # For plotting later
-    regional_gridc = {'lat': np.arange(10.00, 65.15, 0.1), 'lon': np.arange(-98.0, -39.85, 0.1)}
-    nwa_i = xesmf.Regridder(nwa, regional_grid, method='bilinear', unmapped_to_nan=True)(nwa.s_an.isel(depth=0))
-    sw_i = xesmf.Regridder(sw, regional_grid, method='bilinear', unmapped_to_nan=True)(sw.s_an.isel(depth=0))
-    gom_i = xesmf.Regridder(gom, regional_grid, method='bilinear', unmapped_to_nan=True)(gom.s_an.isel(depth=0))
-    combined = xarray.concat([nwa_i, sw_i, gom_i], dim='region').mean('region')
+    regional_gridc = {'lat': np.arange( config['lat']['south']-0.05, config['lat']['north']+0.05, 0.1), 'lon': np.arange(config['lon']['west']-0.05, config['lon']['east']-0.05, 0.1)}
```
andrew-c-ross, line 43: This won't produce the right edge values for some arrays, depending on the start and stop values passed to arange. Instead I think we could modify the `center_to_outer` function already in plot_common to also work with numpy arrays:
```python
def center_to_outer(center, left=None, right=None):
    """
    Given an array of center coordinates, find the edge coordinates,
    including extrapolation for far left and right edge.
    """
    if hasattr(center, 'values'): # handle xarray dataarrays and similar objects
        edges = 0.5 * (center.values[0:-1] + center.values[1:])
    else:
        edges = 0.5 * (center[0:-1] + center[1:])
    if left is None:
        left = edges[0] - (edges[1] - edges[0])
    if right is None:
        right = edges[-1] + (edges[-1] - edges[-2])
    outer = np.hstack([left, edges, right])
    return outer
```

And then replace this line with something like
```python
regional_gridc = {'lat': center_to_outer(regional_grid['lat']), 'lon': center_to_outer(regional_grid['lon'])}
```

Or, since `regional_gridc` is only used once, we could just directly do
```python
grid[1].pcolormesh(center_to_outer(regional_grid['lon']), center_to_outer(regional_grid['lat']), combined, transform=proj, **common)
```

diagnostics/physics/sss_eval.py
```diff
@@ -14,62 +14,54 @@
...
+    combined = combine_regional_climatologies( config, regional_grid)
 
     # Now interpolate to the model grid to compare.
     regional_to_mod = xesmf.Regridder({'lat': combined.lat, 'lon': combined.lon}, model_grid[['geolon', 'geolat']].rename({'geolon': 'lon', 'geolat': 'lat'}), method='bilinear')
     regional_rg = regional_to_mod(combined + 273) 
     regional_rg = regional_rg.where(regional_rg > 0) - 273
     delta_regional = model_ave - regional_rg
```
andrew-c-ross, line 50: This wouldn't have to be in this PR, and I also I should have done this a while ago. We should be able to replace these gymnastics (adding 273, setting anything <=0 to nan because xesmf produces 0s instead of nans, and then subtracting 273 back out) with just `unmapped_to_nan=True`

uwagura:
> Seems like the results are sensitive to what grid the regional climatologies are combined onto. I can reproduce the RMSE = 0.70 by changing the script I used in the paper to use the lat/lon values in this repo's config.yaml

Yeah, I had noticed this previously but forgotten to mention it in this PR. The grid in the config.yaml is based on the grid in the original version of the `sst_eval.py` script. Do we want to keep using this single grid for all scripts, or should we define different grids for different scripts?

yichengt900:
> > Seems like the results are sensitive to what grid the regional climatologies are combined onto. I can reproduce the RMSE = 0.70 by changing the script I used in the paper to use the lat/lon values in this repo's config.yaml
> 
> Yeah, I had noticed this previously but forgotten to mention it in this PR. The grid in the config.yaml is based on the grid in the original version of the `sst_eval.py` script. Do we want to keep using this single grid for all scripts, or should we define different grids for different scripts?

@uwagura, I don’t have a strong preference, but it seems more straightforward to use the same single grid in the example for demonstration purposes. This approach should work unless there are specific accuracy concerns with observational data in certain areas that we’re aware of and want to avoid. @andrew-c-ross, do you have any additional thoughts?

diagnostics/physics/plot_common.py
```diff
@@ -225,3 +233,98 @@ def process_glorys(config, target_grid):
...
+        end_year = 2000 + int(end_year_char)
+    elif end_decade_char == "B":
+        end_year = 2010 + int(end_year_char)
+    elif end_decade_char == "C":
+        end_year = 2020 + int(end_year_char)
+    else:
+        logger.error((f"ERROR: Found unrecognized decade code {end_decade_char} in {clima_file}."
+        "please make sure that decadal climatology follows the National Centers for Environemntal"
+        "Information naming conventions"))
+        raise Exception("Error: unrecognized decade code {end_decade_char} in {clima_file}")
```
yichengt900, line 267: Perhaps we can refactor this part as:
```
    # Map the derived decade and year characters to an actual year, stored as an int
    decade_mapping = {
        "A": 2000,
        "B": 2010,
        "C": 2020
    }

    if end_decade_char in decade_mapping:
        end_year = decade_mapping[end_decade_char] + int(end_year_char)
    else:
        logger.error((f"ERROR: Found unrecognized decade code {end_decade_char} in {clima_file}."
        "please make sure that decadal climatology follows the National Centers for Environemntal"
        "Information naming conventions"))
        raise ValueError(f"Error: Unrecognized decade code {end_decade_char} in {clima_file}")
```
This will make it easier to maintain in the future.

yichengt900 (approved):
@uwagura, thank you for your efforts in generalizing the ss_eval script. The changes look good to me. We plan to extend the script to cover the Arctic and Pacific Islands once we have more information for those domains. Also, the slight difference in time coverage for the NEP data seems acceptable for now, but we can always revisit the issue if concerns arise.

Approved.

andrew-c-ross:
> > > Seems like the results are sensitive to what grid the regional climatologies are combined onto. I can reproduce the RMSE = 0.70 by changing the script I used in the paper to use the lat/lon values in this repo's config.yaml
> > 
> > 
> > Yeah, I had noticed this previously but forgotten to mention it in this PR. The grid in the config.yaml is based on the grid in the original version of the `sst_eval.py` script. Do we want to keep using this single grid for all scripts, or should we define different grids for different scripts?
> 
> @uwagura, I don’t have a strong preference, but it seems more straightforward to use the same single grid in the example for demonstration purposes. This approach should work unless there are specific accuracy concerns with observational data in certain areas that we’re aware of and want to avoid. @andrew-c-ross, do you have any additional thoughts?

Ideally the grid being used for interpolation here should match the grid used by the regional climatologies. The idea was not really to interpolate them to a grid, just to use the interpolation as an easy way of combining different subsets of a larger domain onto the larger domain. For example, for the NWA (and elsewhere?), the regional climatologies are all on a 1/10 degree grid with centerpoints at .05 (ie lat=0.05, 0.15, 0.25, etc). I think it would be best if the grid being interpolated to had the same centerpoints.

diagnostics/physics/plot_common.py
```diff
@@ -225,3 +233,100 @@ def process_glorys(config, target_grid):
...
+
+    logger.info(f"Climatology end year: {end_year}")
+    logger.info(f"End year of last climatology file found successfully.")
+    return end_year
+
+def combine_regional_climatologies(config, regional_grid):
+    """
+    Given a list of domains and a data directory in the config fle, open the list of regional
+    climatology data files, regrid them to the grid defind in regional_grid, and then concat
```
andrew-c-ross, line 278: Typo (defind -> defined)

andrew-c-ross:
To stick with the config files while also adapting to the regional climatology grid, I think it should be possible to do something like setting `np.floor(config['lat']['south'] / .05) * .05` as the start of the lat grid and `np.ceil(config['lat']['north'] / .05) * .05` as the end of the lat grid, and the same for lon. Assuming the other regions are also 1/10 degree with the same centerpoints.

yichengt900:
> To stick with the config files while also adapting to the regional climatology grid, I think it should be possible to do something like setting `np.floor(config['lat']['south'] / .05) * .05` as the start of the lat grid and `np.ceil(config['lat']['north'] / .05) * .05` as the end of the lat grid, and the same for lon. Assuming the other regions are also 1/10 degree with the same centerpoints.

Thanks @andrew-c-ross, I agree that the interpolated grid should have the same centerpoints, and I like the idea you just proposed. @uwagura, do you think you can make one more change to address @andrew-c-ross's point?

uwagura:
> To stick with the config files while also adapting to the regional climatology grid, I think it should be possible to do something like setting `np.floor(config['lat']['south'] / .05) * .05` as the start of the lat grid and `np.ceil(config['lat']['north'] / .05) * .05` as the end of the lat grid, and the same for lon. Assuming the other regions are also 1/10 degree with the same centerpoints.

Won't this result in the same output as what I currently have? I.e`np.floor(0/.05) * .05 = 0.0` and `np.ceil(60/.05) * .05 = 60.0` , so `np.arange( 0.0, 60.0, 0.1) = 0.0, 0.1, 0.2,` and so on.

andrew-c-ross:
Oh, good point. I guess it should be something closer to `round(config['lat']['south'], 1) - .05)` and `round(config['lat']['north'], 1) + .05)`, ie rounding to the nearest 0.1 and then extending by 0.05

yichengt900:
Perhaps something like:

# Grid creation with centerpoints at 0.05 and 0.1 interval
regional_grid = {
    'lat': np.arange(np.floor(config['lat']['south'] * 10) / 10 - 0.05, 
                     np.ceil(config['lat']['north'] * 10) / 10 + 0.1, 0.1),
    'lon': np.arange(np.floor(config['lon']['west'] * 10) / 10 - 0.05, 
                     np.ceil(config['lon']['east'] * 10) / 10 + 0.1, 0.1)
}

uwagura:
@yichengt900 , I've made the change you suggested. The RMSE between the model and the regional climatologies changed ~slightly~ for the nep with the center points grid, but I don't think this is a big issue. 
sss_eval_nep_new_grid

andrew-c-ross (approved):

yichengt900:
@uwagura , Thank you for the updates and for testing the script on NEP as well. I have two quick questions. Below is the SSS plot from Liz's manuscript:

Screenshot 2024-09-24 at 1 58 14 PM

The metrics in her plot differ from ours. I'm wondering if this discrepancy is primarily due to the domain we selected (as it seems we've excluded most of the Bering Sea), or if it's because our combined function applies a weighting method for the overlapping areas, while Liz's approach appears to use the most recent climatology directly. This isn’t a major issue, but I’m curious about which factor contributes more to the differences.

uwagura:
@yichengt900 , It seems to be more sensitve to the domain? I tried to mimic Liz's domain as best as I could ( i.e I extended my coodinates to include the bering sea and excluded everything north of Alaska) and the bias, MEDAE, and correlation moved slightly closer to the values that she has:

sss_nep_Liz_extent

When I remove the weights but keep the same extent, there doesn't seem to be any change

sss_eval_no_weights

Likewise, when I include the Bering Sea and remove the weights, the weights have no effect: 
sss_eval_Liz_extent_no_weights

Do you know if Liz is using the same NCEI climatologies I used?

yichengt900:
@uwagura, thank you for the thorough testing. I believe Liz is using the same data; the major difference seems to be in the overlap area (i.e., above 50°N), where she directly applied the NNP climatology (Version 2). It's good to confirm that the domain is a significant factor, aligning with Andrew's main point. I think this PR is ready to go! Thanks again for your excellent work!

PR #94, opened 2024-09-19
Fixed uv rotation

yichengt900:
We identified a bug in our Python pre-processing tool related to the sign of the UV rotation. The sign was incorrect, so we have reversed it and verified the fix in both the `NEP` and `ARCTIC` domains (see figures below; the black arrow represents true north, the green indicates the previous rotation, and the red shows the corrected rotation). The updated rotation should now accurately map the Earth’s true UV coordinates to the model’s coordinate system.

`NEP`:
nep_test

`ARC`:
arc_test

### DIFF
--- a/tools/boundary/boundary.py
+++ b/tools/boundary/boundary.py
@@ -20,7 +20,7 @@ def rotate_uv(u, v, angle):
     Args:
         u: west-east component of velocity.
         v: south-north component of velocity.
-        angle: angle of rotation from true north to model north.
+        angle: angle of rotation from true north to model north (defined in a counterclockwise direction).
 
     Returns:
         Model-relative west-east and south-north components of velocity.
--- a/tools/boundary/boundary.py
+++ b/tools/boundary/boundary.py
@@ -25,8 +25,8 @@ def rotate_uv(u, v, angle):
     Returns:
         Model-relative west-east and south-north components of velocity.
     """
-    urot = np.cos(angle) * u + np.sin(angle) * v
-    vrot = -np.sin(angle) * u + np.cos(angle) * v
+    urot = np.cos(angle) * u - np.sin(angle) * v
+    vrot = np.sin(angle) * u + np.cos(angle) * v
     return urot, vrot
 
 


theresa-morrison (approved):
The change looks good to me. Thanks for testing this!

theresa-morrison:
I sent Kate an email about this PR, hopefully she is able to confirm if the Arctic OBCs we have been using have this bug.

charliestock:
Thanks for raising and addressing this Yi-Cheng, Andrew and Vimal.  Looking forward to seeing the improvements this yields in the coming weeks.  Perhaps we could include a comment in the code to clarify how the angle is defined in the MOM6 grid?

charliestock:
I think we may still have a bit more work to do on this.  

Consider a domain that is rotated by pi/6 in the counterclockwise direction. (e.g., perhaps one rotated such that the new north-south is aligned with the alongshore direction of the California Current).  If we subject this domain to a truly southward current of -1 along its northern boundary,  in the rotated frame, the old method yields u = -0.5 and v = -0.866.  As expected, the current is still predominantly southward but the rotation has introduced a westward (offshore) flow.  The total magnitude of the current remains 1.  The new method also gives v = -0.866, but gives u = +0.5.  Rather then creating a offshore flow as expected, it creates an inshore flow.

I need to spend some more time with this to unpack all the possibilities and fully convince myself, but the old code seems to give the right answer in the case above.

andrew-c-ross (changes requested):
I think this was an unnecessary panic by me. Thanks @charliestock for showing me the light.

The original formula is for a clockwise rotation of the angle, while the new formula is for a counterclockwise rotation. I thought it was right to switch to the new because angle_dx for our model grids is the counterclockwise rotation relative to true earth. I should have realized that what we actually need to do is coordinate rotation, not vector rotation, in which case the original formula is correct for a counterclockwise rotation of the coordinates. (https://mathworld.wolfram.com/RotationMatrix.html)

You can see this in @yichengt900 's first figure of the NEP. A true north (0, +v) vector should be nearly a (+v, 0) vector on the model coordinates. In the old formula we'd have u ~= sin(angle) * v which yields +v since sin(angle) is positive. In the new formula it would yield -v.

yichengt900:
Thank you, @charliestock and @andrew-c-ross, for your thorough investigation and valuable insights. After reviewing the grid generation source code, I can confirm that the `angle_dx` in the `ocean_hgrid.nc` file represents the orientation of the local coordinate system relative to the spherical Earth, measured counterclockwise. Therefore, if our goal is indeed to perform **coordinate rotation** rather than vector rotation, the original formulation appears to be correct for this purpose.

If either of you could kindly confirm this as well, I will proceed to close the PR to prevent any further confusion for users.

andrew-c-ross:
I talked to Charlie and Matt and we're convinced that we should close this. However, it would be good to add more details to the function docstring so that we remember that the angle_dx is the angle to go counterclockwise from the true earth east to the model east, and that this function is doing a coordinate transformation from true earth coordinates to model coordinates.

PR #95, opened 2024-09-23
Better document how velocity rotation works

andrew-c-ross:
This will hopefully help prevent future cases of #94 by better documenting what the angle rotation function does.


yichengt900 (approved):
Thank you, @andrew-c-ross, for putting this together. The comments look clear and effectively explain the function's purpose. I approve these changes.

PR #97, opened 2024-09-25
Generalized mld_eval.py to use config file

uwagura:
This PR modifies mld_eval.py to use the settings set in the config.yaml file. Given that this script compares model output to global datasets, there were few changes to make. Plots are included for each domain, though due to the lower resolution of DeBoyer and Holte data, some of these plots are not as useful as the others. Still, they are included as a reference.

The only small issue I encountered was with shape of the arctic plot. Despite being projected unto the NorthPolarStereo projection, it still has square boundaries, instead of the round boundaries present in #74. I don't think this is a big issue, I was just confused by the discrepancy and would love other peoples thoughts on it. 

mld_eval_nwa
mld_eval_nep
mld_eval_MHI
mld_eval_arctic

### DIFF
--- a/diagnostics/physics/mld_eval.py
+++ b/diagnostics/physics/mld_eval.py
@@ -4,7 +4,7 @@ and de Boyer 2022.
 Uses whatever model data can be found within the directory pp_root,
 and does not try to match the model and observed time periods.
 How to use:
-python mld_eval.py /archive/acr/fre/NWA/2023_04/NWA12_COBALT_2023_04_kpo4-coastatten-physics/gfdl.ncrc5-intel22-prod/
+python mld_eval.py -p /archive/acr/fre/NWA/2023_04/NWA12_COBALT_2023_04_kpo4-coastatten-physics/gfdl.ncrc5-intel22-prod/ -c config.yaml
 """
 
 import cartopy.crs as ccrs
@@ -14,15 +14,23 @@ from mpl_toolkits.axes_grid1 import AxesGrid
 import numpy as np
 import xarray
 import xesmf
+import logging
 
 from plot_common import annotate_skill, autoextend_colorbar, get_map_norm, open_var
 
+# Configure logging for mld_eval
+logger = logging.getLogger(__name__)
+logging.basicConfig(filename="mld_eval.log", format='%(asctime)s %(levelname)s:%(name)s: %(message)s',level=logging.DEBUG)
 
-def plot_mld_eval(pp_root):
-    model = open_var(pp_root, 'ocean_monthly', 'MLD_003')
-    model_grid = xarray.open_dataset('../data/geography/ocean_static.nc')
+def plot_mld_eval(pp_root, config):
+    model = open_var(pp_root, config['domain'] , 'MLD_003')
+    model_grid = xarray.open_dataset( config['model_grid'] )
     mom_mld_climo = model.groupby('time.month').mean('time').load()
     mom_mld_winter = mom_mld_climo.sel(month=slice(1, 3)).mean('month')
+    logger.info("MODEL_GRID: %s",model_grid)
+    logger.info("MODEL_MLD_CLIMO: %s",mom_mld_climo)
+    logger.info("MODEL_MLD_WINTER: %s",mom_mld_winter)
+    logger.info("Successfully opened model grid, took mean over all months and over winter months")
 
     argo = (
         xarray.open_dataset('/net2/acr/mld/Argo_mixedlayers_monthlyclim_03172021.nc')
@@ -31,9 +39,10 @@ def plot_mld_eval(pp_root):
     )
     argo_grid = dict(lat=argo.lat, lon=argo.lon, lat_b=np.arange(-90, 90.1, 1), lon_b=np.arange(-180, 180.1, 1))
     argo_mld_winter = argo.sel(month=slice(1, 3)).mean('month')
+    logger.info("ARGO_MLD_WINTER: %s",argo_mld_winter)
 
     mom_to_argo = xesmf.Regridder(
-        model_grid[['geolon', 'geolat', 'geolon_c', 'geolat_c']].rename({'geolon': 'lon', 'geolat': 'lat', 'geolon_c': 'lon_b', 'geolat_c': 'lat_b'}),
+        model_grid[ config['rename_map'].keys() ].rename( config['rename_map'] ),
         argo_grid,
         method='conservative_normed',
         unmapped_to_nan=True,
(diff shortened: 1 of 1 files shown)


andrew-c-ross:
@uwagura thanks for working on this. Do you have a config file for the Arctic domain that I could use to test this out?

uwagura:
@andrew-c-ross , the config file is located in `/work/Utheri.Wagura/diagnostic_config_files/config_arctic.yaml`

andrew-c-ross:
@uwagura could you also point me to the /pp/ directory you used for the Arctic plots?

yichengt900:
Hi @uwagura,

Thank you for opening this PR. I'll review your code in more detail soon, but I wanted to offer a quick suggestion about your Arctic plot. The lack of a circular boundary is likely due to the nature of the North Pole projection. Since your Argo/Deboyer dataset includes latitudes below 40N, this is reflected in the plot and impacts the shape.

You can try filtering your data to include only latitudes between 39N and 90N, which should help restore the round shape in your plot. For example:

```
deboyer_mld_winter = deboyer_mld_winter.where((deboyer_mld_winter.lat >= 39) & (deboyer_mld_winter.lat <= 90), drop=True)
```
This might resolve the issue. Let me know how it goes!

andrew-c-ross (approved):
This looks good to me, and I can get it to reproduce the published figure for NWA12 if I just add in weights based on cos(lat) to the skill annotation. I can put in a separate PR to add weighting where it is missing here and in the SST eval script. 

For the handful of Arctic domain plots I've made in the past, I've used `NorthPolarStereo(central_longitude=-110)` for the plot projection. This could also be a separate PR if it still looks good.

diagnostics/physics/mld_eval.py
```diff
@@ -31,16 +39,18 @@ def plot_mld_eval(pp_root):
...
         argo_grid,
         method='conservative_normed',
         unmapped_to_nan=True,
         reuse_weights=False
     )
     mom_rg = mom_to_argo(mom_mld_winter)
     delta = mom_rg - argo_mld_winter
+    logger.info("DELTA: %s",delta)
 
     deboyer = xarray.open_dataset('/net2/acr/deBoyer2022/mld_dr003_ref10m.nc')
```
yichengt900, line 55: Should we consider moving the hard-coded file path to the YAML config? The same applies to the Argo file.

diagnostics/physics/mld_eval.py
```diff
@@ -49,18 +59,28 @@ def plot_mld_eval(pp_root):
 
     # the two observed datasets have the same grid
     delta_deboyer = mom_rg - deboyer_mld_winter
+    logger.info("DEBOYER_MLD_WINTER: %s",deboyer_mld_winter)
+    logger.info("DELTA_DEBOYER: %s",delta_deboyer)
+
+    # Set projection of each grid in the plot
+    # For now, sst_eval.py will only support a projection for the arctic and a projection for all other domains
+    if config['projection_grid'] == 'NorthPolarStereo':
```
yichengt900, line 67: As Andrew suggested, we could add the `central_longitude` parameter here. We could even allow users to choose their preferred value in the YAML configuration.

diagnostics/physics/mld_eval.py
```diff
@@ -49,18 +59,28 @@ def plot_mld_eval(pp_root):
...
         cbar_location='bottom',
         cbar_mode='edge',
         cbar_pad=0.2,
         cbar_size='15%',
         label_mode=''
     )
+    logger.info("Successfully created grid")
 
     levels = np.arange(0, 326, 25)
     bias_levels = np.arange(-50, 51, 10)
```
yichengt900, line 86: I'm fine with the hard-coded levels for now, but we might consider moving them to the YAML configuration in the future. We can let users define different levels for each variable.

diagnostics/physics/mld_eval.py
```diff
@@ -74,39 +94,55 @@ def plot_mld_eval(pp_root):
...
+    logger.info("Successfully plotted DeBoyer data")
 
-    p = grid[2].pcolormesh(argo_grid['lon_b'], argo_grid['lat_b'], delta_deboyer, cmap=bias_cmap, norm=bias_norm)
+    # Model - DeBoyer
+    p = grid[2].pcolormesh(argo_grid['lon_b'], argo_grid['lat_b'], delta_deboyer, cmap=bias_cmap, norm=bias_norm, transform=proj)
     grid[2].set_title('(c) Model - dBM')
     cbar = autoextend_colorbar(grid.cbar_axes[2], p)
     cbar.ax.set_xlabel('Difference (m)')
-    annotate_skill(mom_rg, deboyer_mld_winter, grid[2], dim=['lat', 'lon'], fontsize=8) # TODO: need to cosine weight
+    annotate_skill(mom_rg, deboyer_mld_winter, grid[2], dim=['lat', 'lon'], fontsize=8, x0=config['text_x'], y0=config['text_y'], xint=config['text_xint'], plot_lat=config['plot_lat']) # TODO: need to cosine weight
```
yichengt900, line 125: @andrew-c-ross, do you mean we can do something like this to account for cosine weight?
```
weight = np.cos(np.deg2rad(deboyer_mld_winter.lat))
weight = weight.broadcast_like(deboyer_mld_winter)
```

yichengt900 (commented):
@uwagura, the code looks good to me. I only have a few minor comments. Regarding the question about the boundary shape, you might consider trying my suggestion in another comment.

uwagura:
@yichengt900 It looks like you were right, the only reason why the corners weren't rounded in the MLD plots was because the data is global. When I comment out the `set_extent` function, I get the rounded plot I expected ( see plot below ). I had assumed that `set_extent` would only display data up to the `min_lat`in all cases, but I guess if you have global data and it will display data outside of `min_lat` in `NorthPolarStereo`

Regarding some of the comments in your review: I had thought about adding the Argo/DeBoyer datasets and `central_latitude` parameter to the config file, but I was concerned about the file becoming to crowded with extra arguments, especially as we use it in more scripts. Do you think it's fine if we have a bunch of arguments in the config file that are specific to one or two scripts? I think we can alleviate this problem if we reorganize the config file ( I plan on opening another PR down the road to simply make the config file neater), but I thought I'd get your thoughts first before doing this
mld_eval_arctic_full_extent_rounded_image

yichengt900:
@uwagura, thanks for the updates. I’m glad to hear that my initial thoughts were on the right track. Regarding the arguments and parameters in the config file, I understand and completely agree with your concern. The key principle here is to minimize hard-coding as much as possible. While I’m fine with having script-specific arguments, I agree that we should address this in a separate PR to reorganize the config file—potentially by separating common sections from script-specific sections.

I'm okay with the current codes, but in the next PR, we should try to remove the hard-coded sections.

yichengt900 (approved):

PR #64, opened 2024-07-18
Update NWA-COBALT XMLs for Improved Timing

yichengt900:
As titled, this PR updates runtime parameters in our NWA example XMLs to enhance performance:

1. Change `DT` and `DT_THERM` back to 600s and 1800s, respectively.
2. Use `DT_OBC_SEG_UPDATE_OBGC = 1800`.

The new runtime is approximately 7.53 hours for a 1-year simulation in `production` mode.

PR #99, opened 2024-10-09
Generalize sst_trends

uwagura:
This PR adds the ability to run `sst_trends.py` using the config.yaml file. As always, figures of the new output are attached below, and I wouldn't read too much into any of actual metrics.

I added options to the `process_glorys` function in  `plot_common` to work with script. You can now optionally specify a resample frequency that will be applied to the dataset before take the mean, as well as the whether you want to return the regrid object or the regridded glorys data. These options make the function compatible with this script, but I wasn't sure if the regrid option was good design, since it means that 'process_glorys' now technically has two different return values. I added type hints to reflect this, but I would love to hear people's opinions on this design and whether or not I should revert `process_glorys` to a single return type and avoid calling it in this script. 

Similarly, I avoided calling `process_oisst` in this script because 1.) it regrids model output to the oisst data, while `sst_trends` regrids oisst data to the model output and 2.) it uses a conservative normed method to regrid, while `sst_trends` uses a bilinear method. It's probably possible to add options to `process_oisst` to handle these differences, but it would complicate the logic in `process_oisst` a bit more (especially if I added the two new options from `process_glorys` as well) , so I thought I'd open these changes to discussion as well before implementing them. 

sst_trends_arctic
sst_trends_MHI
sst_trends_nep
sst_trends_nwa

### DIFF
--- a/diagnostics/physics/sst_trends.py
+++ b/diagnostics/physics/sst_trends.py
@@ -10,10 +10,11 @@ import matplotlib.pyplot as plt
 from mpl_toolkits.axes_grid1 import AxesGrid
 import numpy as np
 import xarray
-import xesmf
 import logging
 
-from plot_common import autoextend_colorbar, corners, get_map_norm, annotate_skill, open_var, save_figure, load_config, process_glorys
+from plot_common import( autoextend_colorbar, corners, get_map_norm,
+                         annotate_skill, open_var, save_figure, load_config,
+                         process_glorys, process_oisst)
 
 # Configure logging for sst_eval
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


yichengt900:
@uwagura Thanks for opening this PR! I’ll need to take a closer look later, but the code changes seem solid, and the results remain consistent. @andrew-c-ross, I would greatly appreciate your input, especially regarding the regridding aspect:

In 'sst_eval.py', we regrid the [model to OISST](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/e5a2ec7c57f81c9b5aadbf0a2ad9eff664e72842/diagnostics/physics/plot_common.py#L196-L206) and [GLORYS to the model.](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/e5a2ec7c57f81c9b5aadbf0a2ad9eff664e72842/diagnostics/physics/plot_common.py#L231-L235) Now, in 'sst_trend.py', we’re [regridding both datasets to the model domain](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/e5a2ec7c57f81c9b5aadbf0a2ad9eff664e72842/diagnostics/physics/sst_trends.py#L62-L73). I assume this is to ensure a fair comparison of temperature trends using the same domain. Do you think it would be better to just revert change made in this PR, or perhaps we could adjust the OISST regridding in `sst_eval.py` to maintain consistency across the scripts? Any feedback would be appreciated!

andrew-c-ross:
Re: regridding, my intention (which may not have been implemented in the original version of this script) was to conservatively interpolate the model data onto the observations in all cases where the observations have a resolution of 1/4 deg or coarser. For finer resolution products close to the same resolution as NWA12, I just bilinearly interpolated the observations onto the model grid. For very fine resolution products (only 4 km satellite chlorophyll), I did the same but with conservative interpolation. Thus, the intention was to interpolate the model to OISST, and interpolate GLORYS to the model. I doubt it makes much of a difference, but this was my plan to keep everything consistent.

yichengt900:
Hey @andrew-c-ross, thanks for the input! That clears things up for me. @uwagura, whenever you have a chance, could you update this PR to incorporate Andrew's comments?

uwagura:
@yichengt900 , @andrew-c-ross , I have now updated the sst_trends script to regrid the model data to the oisst data. I had to make a couple of other changes to accommodate this, most notably
1.) `oisst_delta` is now defined as `oisst_delta = mom_rg - oisst_trends`
2.)  The `Model - Oisst` plot now uses `oisst_lonc` and `oisst_latc`
3.)  And I now use `mom_rg` and `oisst_trends` to calculate the skill scores instead of `model_ave` and `oisst_rg`. 

Also, as a result of these changes, it looks like the domain of the `Model  - OISST` plot is now slightly smaller, and the skill metrics are also slightly changed. Let me know if this seems reasonable: 
sst_trends_nwa_mom_to_oisst

uwagura:
Correction to my previous comment: it looks like I forgot to pass in the corner points to the regridder, so it was try to guess the corner points of the target grid, leading to the weird domain in my previous plot. This is fixed now, but the skill scores are still _slightly_ different from the original plot.
sst_trends_nwa_mom_to_oisst_WITH_CORNERS

diagnostics/physics/sst_trends.py
```diff
@@ -25,101 +28,132 @@ def get_3d_trends(x, y):
...
-        xarray.open_dataset('/work/acr/mom6/diagnostics/glorys/glorys_sfc.nc')
-        ['thetao']
-        .sel(time=slice('2005', '2019'))
-        .resample(time='1AS')
-        .mean('time')
-    )
+    # Process OISST and get trend
+    mom_to_oisst, oisst, oisst_lonc, oisst_latc = process_oisst(config, target_grid, model, start =  int(config['start_year']),
+                                                                end = int(config['end_year'])+1, resamp_freq = '1AS',
+                                                                do_regrid = False) # (note that model data is not used if do_regrid = False)
```
yichengt900, line 58: Since we're following Andrew's approach to interpolate the model to OISST and GLORYS to the model, we might want to remove the `do_regrid` option in both `process_oisst` and `process_glorys`. This way, the scripts will just perform the intended interpolations (model to OISST, GLORYS to model), and we can clean up any redundant code (e.g., line #64).

uwagura: I'm just now taking a look at this, and I'm not sure that we can remove the `do_regrid` option from `process_glorys`. This script regrids the `glorys_trend`  to the model grid, and `glorys_trend` is calculated from the `glorys_ave` dataset that is created and returned by this function. 

It looks like the only options are to either 1.) call the `get_3d_trends` function within `process_glorys` which would require adding a command line flag to specify when this function is needed, or 2.) Maintaining the current design that gives you the option to either return the regridder or the regridded dataset. Let me know which sounds better, or if I'm missing another option.

yichengt900: @uwagura, I personally prefer the first option of adding an optional flag in `process_glorys` to allow the function to return trend output when needed, as I think the `do_regrid` option could cause some confusion (since we already know regridding is always required in our case). However, it’s entirely up to you!

yichengt900 (approved):

PR #101, opened 2024-10-18
Split Gulf Steam plot from ssh_eval and rewrite ssh_eval to use config file

uwagura:
This PR separates the gulf stream position and index plot form the mean ssh plot, and generalizes the ssh plot to use the config.yaml file. Plots included below; the arctic and MHI plots were made using only about 2 weeks worth of data, so as always don't read into the metrics or plots too deeply, and please forgive the sloppy formatting!

It looks like the `cartopy.axes.set_xticks` function that `add_ticks` calls does not yet support non-rectangular projections, so as a work around for the arctic I included an if statement before this function is called that checks if the grid is in the `NorthPolarStereo` projection before calling `add_ticks`. This workaround feels a bit clumsy, but should be good for the projections we expect right now. Would appreciate people's thoughts on whether or not there is a better way to do this check. 

mean_ssh_eval_nwa
gulfstream_eval
mean_ssh_eval_arctic
mean_ssh_eval_MHI
mean_ssh_eval_nep

### DIFF
--- a/diagnostics/physics/ssh_eval.py
+++ b/diagnostics/physics/ssh_eval.py
@@ -5,7 +5,7 @@ and a comparison of time mean SSH.
 Uses whatever model data can be found within the directory pp_root,
 and does not try to match the model and observed time periods.
 How to use:
-python ssh_eval.py /archive/acr/fre/NWA/2023_04/NWA12_COBALT_2023_04_kpo4-coastatten-physics/gfdl.ncrc5-intel22-prod
+python ssh_eval.py -p /archive/acr/fre/NWA/2023_04/NWA12_COBALT_2023_04_kpo4-coastatten-physics/gfdl.ncrc5-intel22-prod -c config.yaml
 """
 
 import cartopy.crs as ccrs
@@ -15,10 +15,15 @@ from mpl_toolkits.axes_grid1 import AxesGrid
 import xarray
 import xesmf
 import numpy as np
+import logging
 
-from plot_common import autoextend_colorbar, corners, get_map_norm, open_var, add_ticks, annotate_skill, save_figure
+from plot_common import autoextend_colorbar, corners, get_map_norm, open_var, add_ticks, annotate_skill, save_figure, load_config
 
-def plot_ssh_eval(pp_root, label):
+# Configure logging for ssh_eval
+logger = logging.getLogger(__name__)
+logging.basicConfig(filename="ssh_eval.log", format='%(asctime)s %(levelname)s:%(name)s: %(message)s',level=logging.INFO)
+
+def plot_ssh_eval(pp_root, config, label):
     # Ideally would use SSH, but some diag_tables only saved zos
     try:
         model_ssh = open_var(pp_root, 'ocean_monthly', 'ssh')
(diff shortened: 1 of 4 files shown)


diagnostics/physics/ssh_eval.py
```diff
@@ -1,163 +1,72 @@
...
-    ax.set_title('(c) Gulf Stream index based on SSH variance')
-    ax.set_xlabel('')
-    ax.set_ylim(-3, 3)
-    ax.set_ylabel('Index (positive north)')
-    ax.legend(ncol=4, loc='lower right', frameon=False, fontsize=8)
-
-    save_figure('gulfstream_eval', label=label, pdf=True)
+    # Set projection of each grid in the plot
+    # For now, sst_eval.py will only support a projection for the arctic and a projection for all other domains
+    if config['projection_grid'] == 'NorthPolarStereo':
```
yichengt900, line 50: Not a big deal, but in other scripts (e.g., `sst_eval.py`), you used `p` for the grid projection:
https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/801a17692c9497c6abfa0fbf3db5d63d8629350e/diagnostics/physics/sst_eval.py#L51-L54
It might be a good idea to make the naming consistent with similar scripts. I understand that if you do this, you'll also have to modify the following: https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/6101793cfa12d34269fd8b1ee3d39fe3d233a20e/diagnostics/physics/ssh_eval.py#L95-L96 and https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/6101793cfa12d34269fd8b1ee3d39fe3d233a20e/diagnostics/physics/ssh_eval.py#L102-L103 too.

diagnostics/physics/ssh_eval.py
```diff
@@ -189,29 +108,43 @@ def plot_ssh_eval(pp_root, label):
...
+            # Until support is added, simply skip function if non-rectangular projection is used. Assume
+            # NorthPolarStereo will be the only non-rectangular projection inputted
+            if config['projection_grid'] != 'NorthPolarStereo':
+                add_ticks(ax, xlabelinterval=2, ylabelinterval=2,
+                          xticks=np.arange( config['lon']['west'], config['lon']['east'], 5 ),
+                          yticks=np.arange( config['lat']['south'], config['lat']['north'], 5 ),
+                          projection = proj
+                         )
+            else:
+                logger.warning("WARNING: Cannot set tick marks in non-rectangular projection. Will not call add_ticks")
```
yichengt900, line 130: I’m okay with this approach. If we do want to add ticks to the `NorthPolarStereo` projection plot, we can try using `Gridliner`:

```
        # Add Gridliner for NorthPolarStereo projection
        gridliner = ax.gridlines(draw_labels=True, color='gray', linestyle='--')
        gridliner.xlabels_top = False  # Disable top labels
        gridliner.ylabels_right = False  # Disable right labels
        gridliner.xformatter = ccrs.cartopy.mpl.ticker.LongitudeFormatter()
        gridliner.yformatter = ccrs.cartopy.mpl.ticker.LatitudeFormatter()
        gridliner.xlabel_style = {'size': 10, 'color': 'gray'}
        gridliner.ylabel_style = {'size': 10, 'color': 'gray'}
        gridliner.xlocator = plt.FixedLocator([-180, -90, 0, 90, 180])  # Set longitudes
        gridliner.ylocator = plt.FixedLocator([60, 70, 80, 90])  # Set latitudes
```

yichengt900 (commented):
@uwagura, thanks for the updates! Please also consider renaming line 95 from p to p0:
https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/6101793cfa12d34269fd8b1ee3d39fe3d233a20e/diagnostics/physics/ssh_eval.py#L95

and line 102 from p to p1:
https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/6101793cfa12d34269fd8b1ee3d39fe3d233a20e/diagnostics/physics/ssh_eval.py#L102

Again it's not a big issue, but it would be good to maintain consistency with the other scripts.

Also, please consider moving [plot_gulf_stream.py](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/pull/101/files#diff-68a9106228949d0f186947cf466ff1ff2dbe250296f493d5a37c36edec566c0b) into the `NWA12` subfolder, as it is specific to the `NWA` domain. Make sure to modify the code accordingly to ensure it still works correctly inside the `NWA12` folder.

yichengt900 (commented):
@uwagura, I found a few minor bugs in your PR that were preventing `ssh_eval.py` from running successfully. I’ve fixed these in a new commit—please review it to ensure it aligns with your intentions for this PR. The Gulf Stream script might benefit from additional refinement, but we can address that in a separate PR. I also noticed that the metric differs from the values in the paper, it is due to the different mask we're using. I’m fine with our current approach, and we can revisit this by adding an optional mask to the config file if needed.

uwagura:
@yichengt900 Thanks for the commit. Regarding the changes:

1.) I purposefully edited the gulf stream plot to output figures into the NWA12 directory, because I thought it would be more intuitive for users to find them there than the figures directory in the parent folder. If you think the figures folder is still a better location, we can remove the `output_dir` argument from the `save_figure` plot as well.

2.) I had left a comment in `ssh_eval` about the longitude coordinates in the config file not matching the longitude coordinates range in `model_grid.geolon`, and reminding users to make sure these values align in their own model grid / config files. I decided to leave the responsibility of making sure these values use the same range to the user so that they could make sure all their files were consistent across all scripts. Will all grid files will use the `[-180,180]` range for longitude? if so, we can leave your edit and erase my comment.

yichengt900:
> @yichengt900 Thanks for the commit. Regarding the changes:
> 
> 1.) I purposefully edited the gulf stream plot to output figures into the NWA12 directory, because I thought it would be more intuitive for users to find them there than the figures directory in the parent folder. If you think the figures folder is still a better location, we can remove the `output_dir` argument from the `save_figure` plot as well.
> 
> 2.) I had left a comment in `ssh_eval` about the longitude coordinates in the config file not matching the longitude coordinates range in `model_grid.geolon`, and reminding users to make sure these values align in their own model grid / config files. I decided to leave the responsibility of making sure these values use the same range to the user so that they could make sure all their files were consistent across all scripts. Will all grid files will use the `[-180,180]` range for longitude? if so, we can leave your edit and erase my comment.

@uwagura :
1.) For now, let’s keep all plot outputs in the figures folder. When we have specific plots for each domain, we can revisit and decide on the best organization.

2.) I saw your comment in the code. The issue is that if someone runs the script and config.yaml as-is, they’ll end up with NaN values in the metrics, which isn’t ideal. I’m not fully satisfied with my workaround either, and I don’t have a better solution at the moment—perhaps adding a mask section in the YAML could help. For now, let’s keep your comment, add a TODO note about the current trick we used to revisit later, and address it in a future update.

yichengt900 (approved):

PR #108, opened 2024-11-01
Clean up diag_table files to provide a default CEFI diag_table

yichengt900:
We’re working to provide a CEFI `diag_table` that will offer the CEFI community a standardized set of default model outputs. I've consolidated the three `diag_table` files we use for NWA12 (`diag_table_MOM6_2024-04`, `diag_table_COBALT_2024-04`, and `diag_table.SIS`), removing unused and unregistered variables, along with some COBALT OMIP and budget-check outputs.

The size of this "default" output is 153 GB for NWA12 one year simulation.  Please feel free to review and share any comments or suggestions you might have.


andrew-c-ross:
(1) What do you think about keeping the ocean, sea ice, and BGC diag tables as separate files? That way, for example, physics-only models could drop the BGC diags, and domains like the Pacific Islands could drop the sea ice diags.

(2) This could be a good opportunity to consolidate the different file names within the BGC diag table. Particularly, ocean_cobalt_sfc, ocean_cobalt_omip_sfc, ocean_cobalt_omip_2d should be able to be consolidated into just ocean_cobalt_sfc (with some of the non-surface diags in ocean_cobalt_omip_2d going elsewhere). I think they're separate in part because OMIP wanted certain things in certain ways, but that shouldn't matter for us.

yichengt900:
@andrew-c-ross , thanks for your initial comments.
1) This is a good idea. Originally, @charliestock and I planned to provide a single default diag_table for CEFI, but separating them into `CEFI_MOM6`, `CEFI_SIS`, and `CEFI_COBALT` would allow us to better tailor diag_tables to specific model runs (e.g., physics-only vs. physics + OBGC runs). This would also support the development of additional diag_tables for targeted purposes, such as detailed daily outputs for ocean and BGC components. @charliestock, what are your thoughts?

2) Good point. We don’t need to follow the OMIP protocol for CEFI, and removing OMIP-specific names is a good idea. I’ll work with both you and @charliestock to further consolidate the BGC diag_table.

charliestock:
Hi @yichengt900, @andrew-c-ross,

Thanks for pushing this forward.  I agree with the logic on #1.  I think "CEFI_MOM6_default", "CEFI_SIS_default", and "CEFI_COBALT_default" would be fine. These could then be modified - hopefully gently - by region and product by the regional ocean modeling teams (i.e., CEFI_MOM6_NWA_Hindcast).  

Let's think on #2 a bit more.  I don't like carrying around the extra stuff, but consistency with OMIP does help in some cases.  The broader user community, for example, has set up scripts expecting CMIP names and units.  The challenge is that these CMIP variables don't cover everything we need and aren't always the most natural breakdowns of the BGC.  We may want to ultimately work on an efficient OMIP/CMIP add on to catch a few important fields.  Let's have a first look at the breakdown according to Andrew's suggestion #1 and build from there.

best,
Charlie

yichengt900:
Thanks, @charliestock. Based on the conversation, I’ve split the `diag_table` into `CEFI_MOM6_default`, `CEFI_SIS_default`, and `CEFI_COBALT_default`, and updated the XMLs to use these new diagnostic tables. The history output size for a 1-year `NWA12` run is approximately 150GB.

yichengt900:
@charliestock and @andrew-c-ross, I yhink this PR is ready to go. The configurations for `CEFI_MOM6_default`, `CEFI_COBALT_default`, and `CEFI_SIS_default` are largely aligned with `diag_table_MOM6_2024-04`, `diag_table_COBALT_2024-04`, and `diag_table.SIS`, with some cleanup applied (unused variables have been removed, and each variable now has a brief description). I've kept the OMIP settings in this version, as we can discuss those further in a future update. I plan to merge this PR by the end of the week, so please feel free to reach out if you have any questions or additional feedback.

charliestock (approved):
Hi @yichengt900.  Thank you for putting the defaults together.  I agree that these look good and are clearly moving us in the right direction on the diagnostic front.  Please forge ahead with accepting these as the initial defaults and we can work to refine further in the coming months.

charliestock:
Hi YC, Andrew - yes, please move ahead with establishing these defaults and we can build from there.

Thanks!
Charlie

PR #112, opened 2024-11-12
Separate regional plotting from chl_eval script and add config file to script

uwagura:
This PR generalizes the chl_eval.py script to use the config file in `diagnostics/physics`, and separates its regional plotting capabilities to a separate file that for the moment only applies to the Gulf of Mexico and Northeast US. This script can be generalized to any sub region at a later point in time. 

The script currently relies on the config file in `/diagnostics/physics`, rather than creating it's own. Since this could lead to that config file becoming even more cluttered, do we want to maintaint this design?

The following plots showoutput from the script across the NWA, NEP, and MHI. I currently do not have a `ocean_cobalt_omip_sfc` pp file for the arctic domain, so it is not included in the following plots. 

domain_chl_MHI
domain_chl_nep
domain_chl_nwa


yichengt900:
@uwagura, thanks for opening this PR. I'll take a look later. For now, here are two main points:

1. I agree that having a separate `config.yml` for the BGC diagnostics is likely a better approach.
2. Ensure all plots are saved to the figures folder.

There are some additional issues (e.g., should we allow users to specify their own seasonal definitions?), but we can address those later.

uwagura:
Thanks Yi-Cheng! I addressed your two points above in my most recent commit. Feel free to review this PR whenever you find the time, there is no rush.

yichengt900 (changes requested):
Hi @uwagura,

I conducted a quick test of the new chl_eval scripts. While `chl_eval.py` seems to work as expected, I encountered an issue when attempting to run `NWA12/chl_eval_regional.py`. Below are the steps I followed and the error message I received.

Could you please review this and let me know if there’s anything I might have missed? Additionally, it would be great if you could ensure the sub-regional script functions as intended. Thanks! 
```
module load miniforge 
conda activate /nbhome/role.medgrp/.conda/envs/medpy311/
cd diagnostics/biogeochemistry/NWA12/
python chl_eval_regional.py -p /archive/acr/fre/NWA/2023_04/NWA12_COBALT_2023_04_kpo4-coastatten-physics/gfdl.ncrc5-intel22-prod -c ../config.yaml

Traceback (most recent call last):
  File "/work4/ynt/develop/20241114/CEFI-regional-MOM6/diagnostics/biogeochemistry/NWA12/chl_eval_regional.py", line 27, in <module>
    from plot_common import add_ticks, autoextend_colorbar, corners, annotate_skill, open_var, save_figure, load_config
ModuleNotFoundError: No module named 'plot_common'
```

uwagura:
My bad @yichengt900, I forgot to verify that the regional plotting script still worked after moving it to the NWA12 folder. I think I made all the relevant changes in the most recent commit, let me know if you can run it on your end!

yichengt900 (approved):
@uwagura, thank you for your prompt response! I’ve tested both scripts, and they’re running smoothly without any issues. The changes look good to me. A couple of thoughts for future improvements: we could work on reducing hardcoding in `chl_eval_regional.py` and consider adding an option for users to define their preferred season definitions. These can be tackled in a separate PR in the future. Approved.

PR #113, opened 2024-11-14
Added OM4p25 COBALTv3 example

yichengt900:
Recently, several users have requested access to a global `COBALTv3` configuration. In this PR, we’ve added a 0.25° COBALTv3 example, with ocean physics based on the `OM4p25` configuration and using JRA forcing.

With COBALTv3, users can now set or override default BGC parameters in the `COBALT_input` or `COBALT_override` files, similar to the approach in MOM6/SIS2. Please note that the configurations provided here are still experimental and not yet finalized. However, it should serve as a useful starting point for global cases.  CC @feida6996. @feida6996, it would be appreciated if you could take a quick look at [INPUT/COBALT_input](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/feature/om4p25v3/exps/OM4p25.COBALT/INPUT/COBALT_input) and confirm that it aligns with your parameter choices for a constant river TA and DIC case.


yichengt900:
@feida6996, thank you for your prompt response. The estimate of calcium carbonate flux in the range of 0.4–0.5 Pg-C/yr sounds good to me. As I mentioned in the PR, this is an experimental setup, not the final configuration, so we should be fine. Thanks again!

Issue #123, opened 2024-12-06
Consistency in OBC variable naming convetions


theresa-morrison:
I am using the boundary tools to generate OBC's for the Arctic domain from GLORYS. The options in the config file are great, but there is a difference in the naming convention for OBC's in the Arctic and other domains. For example, the temperature OBC's in the Arctic are `temp_001` but the current scripts make `thetao_001`. These naming conventions also extend to the variables in the OBC files. It looks like`write_glorys_boundary.py` uses the GLORYS names for the files. 

I think we could add a output name option in the config file, or we could convert the Arctic OBC's and xmls to use the names already in `write_glorys_boundary.py`. This issue is not urgent, but I think we should consider this to improve the consistency of our experiments across regions.

theresa-morrison:
@kshedstrom

kshedstrom:
@theresa-morrison It is probably a good idea to rename all the Arctic OBCs to match the rest.

yichengt900:
@theresa-morrison, thank you for opening this issue. I agree that it’s probably better to rename all the Arctic OBCs to match the others. Of course, we could still add a rename option in the BC generation script. @uwagura, could you take a look at `write_glorys_boundary.py` when you get a chance?

uwagura:
Just opened up a PR addressing this. Do we want to add custom name options to the other scripts in this folder, or is it fine if we only have options for `write_glorys_boundary.py` @yichengt900 ?

theresa-morrison:
Thanks for the changes with #124

theresa-morrison:
I think we should leave this PR open as a reminder to change the names of the Arctic OBC variable once OBC's with the GLORYS naming convention are available.

PR #124, opened 2024-12-09
Add option to use non-glorys names in boundary tools

uwagura:
Addresses #123 . This PR adds the a list of output files names to the boundary tool yaml so that you can specify names for the concatenated output other than the glorys variable names


yichengt900:
@uwagura, thank you for opening this PR. I believe it is sufficient to include options only for `write_glorys_boundary.py`, as the OBC files for OBGC contain numerous variables. and it is preferable to maintain consistency by using the same names as those used in the bgc model.

yichengt900 (approved):
@uwagura, thank you for the prompt action. The changes look good, and I have tested them on PPAN. I can confirm that the script renames the OBC files as specified by the user. Approved.

PR #127, opened 2024-12-16
Fix Paths on C5

uwagura:
This PR adds a new variable to all xmls so that that when models are run on c5, they point to t`/gpfs/f5/cefi` instead of `/gpfs/f5/ira-cefi`. I used a new variable as opposed to just using the `GFDL_GROUP` variable for clarity, but if we don't think the new variable is necessary that should be fine as well


yichengt900 (approved):
Changes look, Approved!

PR #137, opened 2025-01-24
Improve timestamp adjustment function in Glorys obc generation workflow

yichengt900:
Several users have reported that the timestamp adjustment function in the current OBC workflow disrupts the original time variables and attributes, and will eventually mess up model run. This PR aims to address the issue by using `decode_times=False` when reading the files. Additionally, we have modified several boundary generation codes to further preserve the original time attributes and encoding.

**_UPDATES_**
I tested the updated workflow with the ARCTIC physics-only case shared by @theresa-morrison, and the new OBC files generated from it ran successfully through the year 1993.

The example obc generation workflow for Arctic can be found on ppan:
`/work/ynt/develop/20250124/new/CEFI-regional-MOM6/tools/boundary/glorys_obc_workflow/mom6_obc_workflow_arctic.sh `

The Arctic 1-month test case can be found here:
`/gpfs/f6/ira-cefi/proj-shared/ARC12_nofk_nsrwsp_newobcs__1.o207690353`

### DIFF
--- a/tools/boundary/boundary.py
+++ b/tools/boundary/boundary.py
@@ -496,7 +496,8 @@ class Segment():
     def regrid_velocity(
                 self, usource, vsource, 
                 method='nearest_s2d', periodic=False, write=True, 
-                flood=False, fill='b', xdim='lon', ydim='lat', zdim='z', rotate=True, **kwargs):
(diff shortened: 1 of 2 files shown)


theresa-morrison (approved):
Approved, Thanks for these changes!

PR #143, opened 2025-02-13
Update Gulf Stream eval with satellite SSH extended through 2023

andrew-c-ross:
This PR updates the pre-computed satellite-based Gulf Stream index and points and the commented out code that shows how they were computed. 

The satellite data is from the same dataset as before but has now been extended through the end of 2023 (the latest available in the dataset). I re-downloaded this dataset and monthly averaged it beforehand. 

Published simulation using old pre-computed satellite data:
gulfstream_eval_old_data

Published simulation using new pre-computed satellite data:
gulfstream_eval_new_data


yichengt900 (approved):
@andrew-c-ross, thank you for updating the pre-computed satellite-based Gulf Stream index and points. Please also provide a reference comparing the old and new versions. Approved.

andrew-c-ross:
> Please also provide a reference comparing the old and new versions.

What sort of reference do you need?

yichengt900:
@andrew-c-ross, thank you for updating the pre-computed satellite-based Gulf Stream index and points. Please also provide a reference comparing the old and new versions. Approved.

> > Please also provide a reference comparing the old and new versions.
> 
> What sort of reference do you need?

It's a typo—please ignore it. Apparently I need more coffee now.

PR #144, opened 2025-02-13
On ppan, use hsmget to copy model files before opening

andrew-c-ross:
Another PR to build towards working with an updated historical simulation. This one adds a wrapper around the `hsmget` tool and uses it in the `open_var` function so that model data is opened from /vftmp instead of /archive. 

The general usage is
```python
hsmget = HSMGet()
ncfile = hsmget(Path('/archive/acr/fre/NWA/2025_01/NWA12_COBALT_2025_01/gfdl.ncrc6-intel23-prod-hdf5/pp/ocean_monthly/ts/monthly/5yr/ocean_monthly.199301-199712.ssh.nc'))
ds = xarray.open_dataset(ncfile) # etc
```

If the above is run on ppan, it will copy the file from archive to
`/ptmp/$USER/archive/acr/fre/NWA/2025_01/NWA12_COBALT_2025_01/gfdl.ncrc6-intel23-prod-hdf5/pp/ocean_monthly/ts/monthly/5yr/ocean_monthly.199301-199712.ssh.nc`

And then from there to 
`/vftmp/$USER/pid####/archive/acr/fre/NWA/2025_01/NWA12_COBALT_2025_01/gfdl.ncrc6-intel23-prod-hdf5/pp/ocean_monthly/ts/monthly/5yr/ocean_monthly.199301-199712.ssh.nc`

Finally it will return the above /vftmp path as a `Path`. 

Future runs of the script from the same ppan session will use the vftmp file immediately. Otherwise, the file will be quickly copied from ptmp as long as that hasn't been swept. 

The biggest advantage of this function would be to work with the raw history files, since frepp will also cache those files in the same place on /ptmp. It's still an improvement to not read the post-processed files directly from archive, though.

If the hsmget module isn't loaded or the script isn't running on ppan, `hsmget()` should just do nothing and directly return the input path(s), allowing them to be opened as usual.

### DIFF
--- a/diagnostics/physics/plot_common.py
+++ b/diagnostics/physics/plot_common.py
@@ -1,10 +1,16 @@
 import cartopy.crs as ccrs
 from cartopy.mpl.ticker import LongitudeFormatter, LatitudeFormatter, LatitudeLocator, LongitudeLocator
+from dataclasses import dataclass
+from getpass import getuser
 from glob import glob
 from matplotlib.colors import BoundaryNorm, ListedColormap
 import matplotlib.pyplot as plt
 import numpy as np
 import os
+from pathlib import Path
+import re
+from subprocess import run, DEVNULL
+from shutil import which
 import xarray
 import xskillscore
 import logging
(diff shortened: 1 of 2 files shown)


diagnostics/physics/plot_common.py
```diff
@@ -14,6 +20,48 @@
...
+        # so that it can be run as a shell command. 
+        esc = re.sub(r'([\(\)])', r'\\\1', cmd)
+        return run(esc, shell=True, check=True, stdout=stdout, stderr=stderr)
+    
+    def _dirs_exist(self):
+        return self.archive.is_dir() and self.ptmp.is_dir() and self.tmp.is_dir()
+
+    def __call__(self, path_or_paths):
+        if which('hsmget') is None or not self._dirs_exist():
+            # If hsmget or /archive, /ptmp, etc are not available, this will just return the input path(s).
```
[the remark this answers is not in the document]
andrew-c-ross: Good idea! I added a reminder to the log message, a comment, and a note in the README.

diagnostics/physics/plot_common.py
```diff
@@ -138,23 +188,32 @@ def add_ticks(ax, xticks=np.arange(-100, -31, 1), yticks=np.arange(2, 61, 1), xl
...
-        files = glob(os.path.join(pp_root, 'pp', kind, 'ts', freq, '1yr', f'{kind}.*.{var}.nc'))
-        os.system(f'dmget {" ".join(files)}')
-        return xarray.open_mfdataset(files)[var]
-    elif len(glob(os.path.join(pp_root, 'pp', kind, 'ts', freq, '5yr', f'{kind}.*.{var}.nc'))) > 0:
-        files = glob(os.path.join(pp_root, 'pp', kind, 'ts', freq, '5yr', f'{kind}.*.{var}.nc'))
-        os.system(f'dmget {" ".join(files)}')
-        return xarray.open_mfdataset(files)[var]
+    # For now, replicate previous behavior by looking for a single file covering 1993--2019,
+    # and if not found then looking for multiple files written in either yearly or 5-yearly chunks.
+    # Later this can be replaced to more intelligently look for the right file(s).
```
yichengt900, line 198: I like this comment `intelligently look for the right file(s)` : )

andrew-c-ross: I have a PR ready for once this one is merged :smiley:

PR #145, opened 2025-02-13
More intelligently search for the right post-processed files

andrew-c-ross:
To support newer simulations, this changes the behavior of `open_var` to look for the right post-processed files, without any hard coding of the dates or chunk lengths. 

The general assumptions are
1. A path following the standard frepp naming, for example `{pp_root}/pp/ocean_monthly/ts/monthly/`, exists (error if not)
2. Within that path, one or more directories with names like `*yr` exist (error if not)
3. Of the directories above, whichever has the largest number in the name and contains one or more post-processed files of the right variable is the one to use. 

We generally post-process two different chunk sizes, one of 5 years and the other the length of the simulation. This should work for that. One way to get a suboptimal result would be if the simulation length was longer than the longest chunk; for example, a 25 year simulation with 5 year and 20 year chunks. This is an unlikely user error though.

### DIFF
--- a/diagnostics/physics/plot_common.py
+++ b/diagnostics/physics/plot_common.py
@@ -1,6 +1,7 @@
 import cartopy.crs as ccrs
 from cartopy.mpl.ticker import LongitudeFormatter, LatitudeFormatter, LatitudeLocator, LongitudeLocator
 from dataclasses import dataclass
+import errno
 from getpass import getuser
 from glob import glob
 from matplotlib.colors import BoundaryNorm, ListedColormap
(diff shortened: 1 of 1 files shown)


diagnostics/physics/plot_common.py
```diff
@@ -192,28 +193,27 @@ def open_var(pp_root, kind, var, hsmget=HSMGet()):
...
-        for chunk in possible_chunks:
-            short_dir = pp_root / 'pp' / kind / 'ts' / freq / f'{chunk}yr'
-            if short_dir.is_dir():
-                break
-        else:
-            raise Exception('Did not find directory for postprocessed files')
-        short_files = list(short_dir.glob(f'{kind}.*.{var}.nc'))
-        if len(short_files) > 0:
-            tmpfiles = hsmget(sorted(short_files))
+    if not (pp_root / 'pp' / kind / 'ts' / freq).is_dir():
```
yichengt900, line 196: Perhaps we can construct the post-process path first to avoid repetition?
`post_process_path = pp_root / 'pp' / kind / 'ts' / freq`

uwagura (approved):
ran it successfully with my own 27 year run, looks good to me!

diagnostics/physics/plot_common.py
```diff
@@ -192,28 +193,27 @@ def open_var(pp_root, kind, var, hsmget=HSMGet()):
...
+        # Look through the available chunks and return for the 
+        # largest chunk that has file(s). 
+        matching_files = list(chunk.glob(f'{kind}.*.{var}.nc'))
+        # Treat 1 and > 1 files separately, though the > 1 case could probably handle both. 
+        if len(matching_files) > 1:
+            tmpfiles = hsmget(sorted(matching_files))
             return xarray.open_mfdataset(tmpfiles, decode_timedelta=True)[var] # Avoid FutureWarning about decode_timedelta
-        else:
-            raise Exception('Did not find postprocessed files')
+        elif len(matching_files) == 1:
```
yichengt900, line 213: It looks like we may not need this check since matching_files will either be multiple files or a single file in our case?

andrew-c-ross: My intent was to do nothing and go to the next chunk directory if no matching files were found in the current chunk directory (for example, if frepp failed or is still running, one could have 5yr/ and 30yr/ but no files in 30yr/). Does the logic work for that case?

yichengt900:
Thanks again, @andrew-c-ross, for your contribution! Just being a little picky about tiny details, but everything looks perfect to me!

diagnostics/physics/plot_common.py
yichengt900: Good point. In that case, we should keep both checks so the code will work in that scenario.

PR #146, opened 2025-02-14
Add NEP Figures

uwagura:
This PR adds Jupyter Notebooks for generating the Figures used in Liz's NEP paper. All paths have been modified to work with files stored on the role accounts work directory, so the Python notebooks should work out of the box once you activate our python environment (i.e the one stored in `/nbhome/role.medgrp/.conda/envs/medpy311`). 

As noted in the README, there is one notebook written in R that currently does not work because of an ongoing conflict between the libraries it uses and a Jupyter based R install. This seems to be a larger issue - see the discussion [here](https://github.com/afsc-gap-products/akgfmaps/issues/145) for the latest details. In any case, I've decided to just leave a warning in the README for now that the notebooks doesn't work, and updat the README once we have a working R environment.


uwagura:
These scripts are now available in the CEFI github: https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/NEP-Figures

yichengt900:
Since we have the public github NEP-Figures repo [here](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/NEP-Figures), will close this draft PR now.

PR #153, opened 2025-03-10
create var extract script in postprocessing

chiaweh2:
The PR is meant for adding a postprocesing script under xmls folder. The postprocessing script's goal is to xtracts a single variable or multiple variables from history tar files generated by a model FRE run. The extracted variables are stored in a specified output directory for further analysis.

The PR includes
1. add a postprocessing directory under xmls
2. add a python script on extracting the variable from history files created from the model output
3. add a json file that the python script read for important information related to file location (input & output) and variable that need to be extracted.


chiaweh2:
@yichengt900 please feel free to let me know if there is more information needed in the docstring or need to be separated out as a single markdown file.

xmls/postprocessing/extract_var_from_history.py
```diff
@@ -0,0 +1,208 @@
...
+        if os.path.isfile(output_var_filename):
+            logging.warning(
+                'extract variable %s exist',output_var_filename
+            )
+        else:
+            logging.info(
+                'extracting variable %s ...',output_var_filename
+            )
+            subprocess.run(
+                ["ncks", "-O", "-h", "-v", var, output_filename, output_var_filename], check=True
```
yichengt900, line 132: Before running `ncks`, check if the tool is available. If it’s not, log an error and exit the script.

xmls/postprocessing/extract_var_from_history.py
```diff
@@ -0,0 +1,208 @@
...
+        # remove individual files
+        for file in allfiles:
+            os.remove(file)
+        logging.info('remove individual files')
+
+    return
+
+if __name__ == "__main__":
+    if len(sys.argv) != 2:
+        print("Usage: python extract_nc.py <config_file.json>")
```
yichengt900, line 169: again should be `[extract_var_from_history.py`

yichengt900 (commented):
Hi @chiaweh2, thanks for opening this PR and contributing the script! I have tested it on PPAN, and it worked as expected. The script looks good, and I only have a few tiny comments on it.

xmls/postprocessing/extract_var_from_history.py
```diff
@@ -0,0 +1,208 @@
...
+- Supports extracting multiple variables at once.
+
+### Requirements:
+- The script expects a configuration file (`example_extract_var.json`) as input.
+- The `example_extract_var.json` file must specify the directories, archive naming conventions, 
+  variable names, and the range of years to process.
+
+### Example Usage:
+```bash
+python extract_nc.py example_extract_var.json
```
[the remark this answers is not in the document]
chiaweh2: Agreed! should the module load be added as docstring or is this should be included as a separate bash shell scrip to be able to run the python script?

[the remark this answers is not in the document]
yichengt900: I think, for now, we can leave it in the docstring so that users running it on PPAN have a quick reference guide

chiaweh2:
Thanks you for the feedback and testing. I have addressed the comments and test a few different cases.

xmls/postprocessing/extract_var_from_history.py
```diff
@@ -0,0 +1,227 @@
...
+                logging.warning(
+                    'extracted tar file %s exist, skipping...',output_filename
+                )
+            else:
+                logging.info(
+                    '(tar -xf) extracting tar file %s ...',output_filename
+                )
+                # untar single file based on subexp_name
+                subprocess.run(
+                    ["tar", "-xf", tar_path, "-C", Path(output_dir), './'+file_to_extract],
```
uwagura, line 132: @yichengt900 , since the tar files are on archive, do we need to call `dmget` on it first?

xmls/postprocessing/extract_var_from_history.py
```diff
@@ -0,0 +1,227 @@
...
+        config = load_config(config_file)
+        output_directory = config["output_directory"]
+        archive_directory = config["archive_directory"]
+        archive_subexp_name = config["archive_subexp_name"]
+        variable_names = config["variable_names"]
+        start_yr = config["start_year"]
+        end_yr = config["end_year"]
+
+        # run extract variable
+        for variable in variable_names:
```
uwagura, line 214: This is a minor point, but it maybe help to add print statements as we loop through the variables and the years so that users can keep track of where they are in the process and make sure it's actually progressing and not frozen.

xmls/postprocessing/extract_var_from_history.py
uwagura: and on that point, we should probably either stage the files on `vftmp` while extracting variables or make sure that `output_dir` is not a location on archive so that we don't run code there

chiaweh2: @uwagura the dmget is a good question. I notice the dmget can exist or not exist for the following command to work. Does that mean the dmget is implicitly applied if not specified?

chiaweh2: @uwagura I agree to give some instruction on the output dir but it seems a bit out of the scope if we force the output dir in a specific location? Or maybe not?  I am happy to implement it if this is required by the data policy.

uwagura: I think `dmget` might be called implicitly on files that are on tape, I've just never tried it before. I just know GFDL Operations prefers that we call it on all files on archive before using them, which is why I suggested this.

uwagura: > @uwagura I agree to give some instruction on the output dir but it seems a bit out of the scope if we force the output dir in a specific location? 

Oh to clarify, I was thinking we could 1.) untar the `/archive` to`vftmp` 2.) extract the variables on `vftmp`  then 3.) move the extracted files to`output_dir`, so user's still have flexibility on what ouput dir they choose. I think this is the recommend way to work with files on `archive`, but others may know better!

xmls/postprocessing/extract_var_from_history.py
chiaweh2: good point. I can add this.

xmls/postprocessing/extract_var_from_history.py
chiaweh2: ah! I see. that make sense. I am not familiar with the vftmp but based on the [wiki](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/wiki/Analyzing-Model-Output). It seems ideal for the possibly large netCDF untared and will be swept if left there for too long? Do you think letting user specified this in the json file is better or letting the script creating a temp folder in vftmp is better?

chiaweh2: I will add the dmget to make sure the script is operated following the operational preference!

yichengt900: > ah! I see. that make sense. I am not familiar with the vftmp but based on the [wiki](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/wiki/Analyzing-Model-Output). It seems ideal for the possibly large netCDF untared and will be swept if left there for too long? Do you think letting user specified this in the json file is better or letting the script creating a temp folder in vftmp is better?

@chiaweh2, normally, when we submit a Slurm job on PPAN, it assigns an environment variable like `TMPDIR=/vftmp/username/pid$JOB_ID`, where all intermediate work can be done. I think in our case, we can have the script create a temporary folder and clean it up at the end?

andrew-c-ross: FWIW `hsmget` (possibly using my wrapper [here](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/548e7be474b03dfb696a786ff2779b219a133031/diagnostics/physics/plot_common.py#L28)) could be used to handle the copying from archive to a tmp dir. It also copies to the longer term /ptmp as an intermediate step. This is nice if frepp has already run on the model history, since it will leave the files cached in /ptmp and you might not need to get the from /archive.

chiaweh2: @yichengt900 I currently did not clean up the directory since original intension is to let the script run check to see if the file is in vftmp if a different execution happen to need the same files. The potential benefit is to avoid untaring files that has previously already untared in the vftmp folder. This would potentially save some time especially related to daily file untar. but I am happy to implement vftmp clean up if this is a better practice.

chiaweh2: @andrew-c-ross does that mean the untar (or dmget?) file could be already cached in /ptmp from previous frepp run? so we could potentially avoid the untar (or dmget?) to be performed again? I am checking out the hsmget -q -a {self.archive} -w {self.tmp} -p {self.ptmp} line in the code and does this create a copy tar out of archive on both vftmp and ptmp? Is this a better practice than just dmget in archive?

yichengt900: @chiaweh2 As @andrew-c-ross mentioned in his original comment, the command `hsmget -q -a {self.archive} -w {self.tmp} -p {self.ptmp}` first attempts to copy the target file from `archive` to `ptmp` and then to `vftmp` if it is not found in either location. It then returns the `vftmp` path containing the target file.

My first impression is that this approach avoids redundant copies of the tar file if it already exists in `vftmp` (or speeds up access if it's already in `ptmp`, since copying from `ptmp` to `vftmp` sould be relatively fast) within the same ppan session.

On second thought, I think it could be a useful addition (though I believe the current script is sufficient, so it's totally up to you) to utilize this wrapper in your script, eliminating the need to specify a `vftmp` folder in json. Implementation should be straightforward (see the example below), but let me know if you need any help.
```
sys.path.append("../../diagnostics/physics")
from plot_common import HSMGet
hsmget = HSMGet()
archive = hsmget(tar_path)

# The new archive should be the vftmp path to the target tar file.
# You can then use it for any further actions.

```

chiaweh2: @andrew-c-ross @yichengt900 thank you both for the explanation! This is very helpful. I tried implement it but the class `[HSMGet](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/548e7be474b03dfb696a786ff2779b219a133031/diagnostics/physics/plot_common.py#L28)` is in the script with many import which seems to take some time and also seems to be missing xskillscore when standard module load python/3.11 . Do you have any recommendation on this? We can of course just copy the entire class to the current script but I feel this is rather not so elegant and will easily may the two class to diverge if further changes is needed in the future.

andrew-c-ross: I'll defer to @yichengt900 here but I do think it's sounding like it would be simpler to merge this PR as is now and think about using hsmget some other time, perhaps with a PR to move hsmget and some other tools to a different place. 

If you want to continue experimenting, though, you should be able to use the conda environment that our GFDL group maintains:
```sh
module load miniforge
conda activate /nbhome/role.medgrp/.conda/envs/medpy311
```

yichengt900 (approved):

xmls/postprocessing/extract_var_from_history.py
chiaweh2: sounds good! thank you all for the suggestion and review.

chiaweh2:
@andrew-c-ross @yichengt900 @uwagura Okay, I just found one problem myself. As the script is originally designed to extract history generated from hindcast. I was not thinking about forecast and reforecast history file format. The tar file is partially hard coded as {year}0101.nc.tar. But I noticed this is not the case for forecast and reforecast. I need to modified to be more generalized so it also work on future experiment. I will push a new edit soon on this. Sorry for the delay.

PR #155, opened 2025-03-18
updating README to indicate NEP-specific file for WOA climatological sst

amoebaliz:
The SST values used for calculating oxygen saturation are from a domain-specific climatology from World Ocean Atlas. The current file is for the NWA. We need to update this with sst values regridded to the NEP domain.


yichengt900:
@amoebaliz, thanks for opening this PR! Could you point me to the location of `woa_nep5k_sst_climo.nc` on PPAN so I can have @uwagura give it a try? Thanks!

yichengt900 (approved):
@amoebaliz, thanks for those changes! Good catch on the WOA files. I also took this opportunity to update the woa filename and correct the path to include the necessary `nc64startup` package in the matlab script. Approved.

Next, we’ll work on updating the MATLAB scripts to make them consistent with what you’ve been doing for the NEP. Thanks again!

PR #169, opened 2025-04-08
Modify longitude handling in subset template to handle domains that cross the dateline

gabyneg:
When a domain crosses the dateline (in my case the Pacific Islands regional model), generating the glorys bgc with the current workflow might not be as straight forward. Therefore, converting the longitude coordinates from a -180 to 180 system to a 0 to 360 grid while maintaining the full global coverage, makes it easier to select the desired domain without interruption.

### DIFF
--- a/tools/boundary/glorys_obc_workflow/mom6_obc_workflow.sh
+++ b/tools/boundary/glorys_obc_workflow/mom6_obc_workflow.sh
@@ -90,7 +90,7 @@ log_message "Generating config.yaml..."
 cat <<EOF > config.yaml
 _WALLTIME: "1440"
 _NPROC: "1"
-_EMAIL_NOTIFACTION: "fail"
+_EMAIL_NOTIFICATION: "fail"
 _USER_EMAIL: "$USER@noaa.gov"
 _LOG_PATH: "./log/$CURRENT_DATE/%x.o%j"
 _UDA_GLORYS_DIR: "/uda/Global_Ocean_Physics_Reanalysis/global/daily"
@@ -157,7 +157,7 @@ if ! $DO_NCRCAT; then
     	    subset_args="$subset_args --convert_lon"
 	fi
         subset_job_id=$(sbatch --job-name="glorys_subset_${year}_${month}_${day}" \
-                              scripts/subset_glorys.sh $year $month $day | awk '{print $4}')
+                              scripts/subset_glorys.sh $subset_arags | awk '{print $4}')
 
         log_message "Submitting fill_nan job for $current_date..."
         fill_job_id=$(sbatch --dependency=afterok:$subset_job_id \
--- a/tools/boundary/glorys_obc_workflow/mom6_obc_workflow.sh
+++ b/tools/boundary/glorys_obc_workflow/mom6_obc_workflow.sh
@@ -9,10 +9,11 @@ set -eu
 
 # Helper functions
 print_usage() {
-    echo "Usage: $0 START_DATE END_DATE [--ncrcat] [--adjust-timestamps]"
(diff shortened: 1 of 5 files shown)


tools/boundary/glorys_obc_workflow/template/subset_glorys_template.sh
```diff
@@ -76,9 +76,11 @@ process_variable() {
         if [[ -f $filename ]]; then
             echo "Processing file: $filename"
             local subset_file="$var_dir/${var}_subset.nc"
-
+            local tmp_file="$var_dir/${var}_tmp.nc"
+            cdo sellonlatbox,0,360,-90,90 $filename $tmp_file
```
yichengt900, line 80: @gabyneg, I like the idea you proposed to include this conversion as an option. I would suggest improving this part as follows:

In the Input arguments, consider adding an optional 4th argument after the first 3 arguments
```
# Initialize convert_lon to false
convert_lon=false

# Check remaining arguments for --convert_lon
shift 3  # Shift past the first 3 arguments (year, month, day)

for arg in "$@"; do
    if [[ "$arg" == "--convert_lon" ]]; then
        convert_lon=true
        break
    fi
done
```

Then we can improve this line to:
```
# Check if convert_lon is true
if [[ "$convert_lon" == "true" ]]; then
    echo "Running longitude conversion with cdo sellonlatbox..."
    
    # Check if cdo is available
    if ! command -v cdo &> /dev/null; then
        echo "Error: cdo command not found."
        exit 1
    fi

    # Run the cdo longitude conversion
    if ! cdo sellonlatbox,0,360,-90,90 "$filename" "$tmp_file"; then
        echo "Error: cdo sellonlatbox failed for $filename"
        exit 1
    fi
else
    # If no conversion needed, copy the file
    cp "$filename" "$tmp_file"
fi
```
We probably also want to modify the main control script to allow the user to enable the convert_lon option, but I can handle that part if you're too busy.

gabyneg: Thank you @yichengt900 , I have just included those changes, and I will work on including the convert_lon option in the main control script, but I will let you know if I have any trouble doing so

tools/boundary/glorys_obc_workflow/mom6_obc_workflow.sh
```diff
@@ -147,6 +152,10 @@ if ! $DO_NCRCAT; then
         day=$(date -d "$current_date" +%d)
 
         log_message "Submitting subset job for $current_date..."
+	subset_args="$year $month $day"
```
yichengt900, line 155: See my comment [here](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/commit/fa89215741a389f7985bcec576a99b0949756c1b#r155026641)

tools/boundary/glorys_obc_workflow/template/subset_glorys_template.sh
```diff
@@ -2,23 +2,37 @@
 #SBATCH --partition=batch
 #SBATCH --time={{ _WALLTIME }}
 #SBATCH --ntasks={{ _NPROC }}
-#SBATCH --mail-type={{ _EMAIL_NOTIFACTION }}
+#SBATCH --mail-type={{ _EMAIL_NOTIFICATION }}
```
yichengt900, line 5: Thanks for catching the typo. Unfortunately, if you change this one, you'll also need to update it in the main script as well as all the other template scripts. :)

gabyneg: got it! Thank you for catching that, I believe I have updated it now in the main script and the rest of the template scripts

yichengt900 (approved):
@gabyneg, thanks for your prompt action! The code changes look good to me now—approved!

Issue #28, opened 2024-04-25
XML checkouts are slow. Is it worth shallow cloning?


andrew-c-ross:
Edit: going to close this because I misread the time command output.

yichengt900:
@andrew-c-ross, jsut let you know that there are also some ongoing discussions in GHAG regarding the performance issues of Gaea f5. We've noticed that the compiling time as well as the time of data transfer are extremely unstable and need to be addressed. Just for your information

andrew-c-ross:
Sounds good. I've definitely noticed the long transfer times from Gaea to GFDL. Mentioned this to Dennis yesterday since I didn't know if it was a local or remote issue.

PR #182, opened 2025-05-19
Updated mom6 tag

yichengt900:
As titled. This is a regualr update for ~mom6 and ocean_BGC tags~ mom6 tag only for now.

[This PR ](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/pull/135)from ocean_BGC changes the baseline in NWA and NEP due to the use of a real (floating-point) exponent instead of the original integer exponent, resulting in subtle differences.


yichengt900:
Unfortunately, the new MOM6 tag (synced with the current dev/gfdl MOM6) produced the following error message in the NEP domain when using the same mask table we used for the production run, `mask_table.524.32x80`:
```
FATAL from PE  1024: MPP_DEFINE_DOMAINS(mpp_compute_extent): domain extents must be positive definite.


FATAL from PE  1016: MPP_DEFINE_DOMAINS(mpp_compute_extent): domain extents must be positive definite.


FATAL from PE   955: MPP_DEFINE_DOMAINS(mpp_compute_extent): domain extents must be positive definite.


FATAL from PE   963: MPP_DEFINE_DOMAINS(mpp_compute_extent): domain extents must be positive definite.
```

I will put this PR on hold and wait for further investigation.

CC @theresa-cordero

yichengt900:
OK, further investigation showed that the following section of code is causing the issue we encountered :
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/blob/5f6a7600006a53998104aff3d13410de4c6206c8/src/framework/MOM_domains.F90#L446-L451

This new code attempts to create an unmasked domain and write out unmasked ocean geometry, even if a mask_table is used. For some reason, it doesn't work with the NEP10k 43x80 mask_table. Currently, this section of code is always executed when a mask_table is specified.

Would it be possible to add a runtime parameter to control this behavior? That way, we could disable this section if we don’t want to generate an unmasked ocean geometry file.

theresa-cordero:
We could add a parameter to stop this. Should I open a PR to dev/cefi?

yichengt900:
Sure, that would be great! I can merge it first, then test the latest OBC PR that was just merged yesterday.

andrew-c-ross:
This same error popped up in older versions of MOM6 with certain layouts of the NWA25 model. I could never figure out why it was happening, but if/when there's a tentative fix I'd be happy to test it out

theresa-cordero:
I encountered this problem with the NWA12 with the 50x50 mask table, although Yi-Cheng did not. 

I added a flag to avoid this code in the MOM6 dev/cefi [PR #28](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/pull/28)

It will be worth knowing why these domain decompositions are causing problems without mask tables.

PR #190, opened 2025-05-29
Updated CI build and run driver

yichengt900:
As titled. Users can now run the CI process for NWA and NEP to test their code updates.


theresa-cordero:
If including `mosaic` in the list paths does not cause a fatal error then I think it should not be removed so that users can run with older versions of FMS more easily if needed.

yichengt900:
@theresa-cordero , thanks. I can change it back once the CI testing is done.

Issue #198, opened 2025-06-09
Update CEFI YAML configurations


singhd789:
There has been some more yaml development since the MSD CMIP7 sprint. In order for yamls to work successfully, some updates are needed. 

*In `yamls/NWA12/yaml_include/settings.yaml`*:
- [ ] `pp_chunk_a` and `pp_chunk_b` --> replaced with `pp_chunks` like so:
```
 pp_chunks: 
    - *PP_CHUNK_A
    - *PP_CHUNK_B
```
- [ ] add to `switches` if needed?:
```
do_refinediag: False
do_atmos_plevel_masking: False 
do_preanalysis: False
do_analysis_only: False
```

*In `yamls/NWA12/platforms.yaml`*:
- [ ] include this platform: https://github.com/NOAA-GFDL/fre-examples/blob/main/platforms.yaml#L21-L31 --> strips out intel compiler and makes end result container shareable and smaller

*In `yamls/NWA12/CEFI_NWA12_cobalt.yaml`*:
- [ ] can probably update `&FMS_GIT_TAG` --> `"2025.02.01"` 

**Steps to test** (example experiment):
```
module load fre/2025.03

git clone git@github.com:NOAA-GFDL/CEFI-regional-MOM6.git
cd CEFI-regional-MOM6/yamls/NWA12

# creates ~/cylc-src/MOM6_SIS2_GENERIC_4P_compile_symm__ppan__debug
fre pp checkout -e MOM6_SIS2_GENERIC_4P_compile_symm -p ppan -t debug

# reads/parses yamls to create "combined yaml" and validates against fre_pp.json schema
fre pp configure-yaml -e MOM6_SIS2_GENERIC_4P_compile_symm -p ppan -t debug -y CEFI_NWA12_cobalt.yaml
```
That step usually errors out if there are formatting issues in the yaml because it would not be valid

....

(usually we would do `fre pp validate ...` before installing, but it is currently showing keys/values it deems as invalid, when really they have been removed from the configs and don't matter here)
```
fre pp install -e c96L65_am5f8d6r3_amip -p ppan -t debug
fre pp run -e c96L65_am5f8d6r3_amip -p ppan -t debug 
```

uwagura:
Addressed by #217

PR #217, opened 2025-07-09
Updated yamls to address tasks in issue #198

uwagura:
Address issues raised in #198


singhd789 (approved):
Looks good! Thank you!

There might be some more changes happening soon so stay tuned!

PR #221, opened 2025-07-30
Auto-update submodules to latest dev/cefi (auto/update-submodules-20250730)

yichengt900:
This PR automatically updates the specified submodules to the latest commit on their `dev/cefi` branch.

**Updated Submodules:**

- src/MOM6 (updated to commit 18e8cb3b19bd4d587a06dd27faa0aaed48f7e45b)

---
Auto-generated by GitHub Actions.


theresa-cordero:
The update to MOM6 has caused problems with the ocean_bgc code. I think this is because of the changes in MOM_diagnose_MLD.F90 from the commit `Add ePBL BBL mstar and bottom mixed layer diagnostics` number 32849fa. These changes conflict with the call to `diagnoseMLDbyEnergy` in MOM_generic_tracer.F90 so the code will not compile. 
I was able to fix this by modifying  MOM_generic_tracer.F90 lines 777-778:
```
      call diagnoseMLDbyEnergy((/-1, -1, -1/), h_old, tv, G, GV, US, (/CS%mld_pha_deng, &
              CS%mld_pha_deng, CS%mld_pha_deng/), (/1,nk/), CS%diag, MLD_out=mld_pha)

```
I will submit a PR with this change to ocean_BGC then I think this change could be accepted.

theresa-cordero:
See ocean_bgc [PR 157](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/pull/157)

PR #223, opened 2025-08-07
Add script for merging historical and ssp245 co2 forcing

andrew-c-ross:
This adds a script which combines the historical (1915--2014) and ssp245 (2015--2064) atmospheric co2 forcing together into one file and writes the result in a way that can be read by FMS. This script is a modernized version of the original that generated the file we've been using for a while.

The result of this script currently lives on Gaea at /gpfs/f6/ira-cefi/proj-shared/global/co2/mole_fraction_of_co2_extended_ssp245_v2.nc. Ideally it should go to the same location in world-shared, which is writable by @yichengt900 . If the file can be copied there I will send a PR to update the XMLs too. 

The co2 forcing file generated by this script will likely slightly change BGC answers starting in 2015 due to using the finely-interpolated version of the file for the scenario. This file may not have been available when I first made this forcing data. An answer change before 2014 is possible due to slight changes in the time and time_bnds.


andrew-c-ross:
cc @amoebaliz

andrew-c-ross:
@amoebaliz pointed out that the SSP forcing starting in 2015 is using a noleap calendar. What I do here, and did in the previous version, is ignore this and overwrite the times with a standard gregorian calendar. I estimated that this does not cause any drift in the forcing but may result in a difference in the applied forcing of about +-1% of the annual cycle amplitude. For atmospheric CO2 this should be negligible.

@uwagura could you take a quick look at this and the related PRs and merge if they look good?

uwagura (approved):
Was able to run script successfully, everything looks good!

uwagura:
are we ready to merge this @andrew-c-ross

PR #238, opened 2025-09-15
Set MOM parameters that were using defaults in NEP experiment

theresa-cordero:
With changes to the defaults in MOM6, some parameters that were not set explicitly in the NEP experiment now must be set to preserve old solutions.

These parameters are:
```
TIDES_ANSWER_DATE = 20230630
WAVE_INTERFACE_ANSWER_DATE = 20221231 
MASS_WEIGHT_IN_PRESSURE_GRADIENT_TOP = False
```


uwagura (approved):
As discussed earlier, we should also update the main xml for this configuration to use the latest settings, but otherwise I was able to reproduce old answers with these changes.

PR #242, opened 2025-09-19
Update MOM6 and ocean_BGC to point to the latest commits

uwagura:
This PR points the submodules to their latest versions.


uwagura:
tests initially failed because my branch originally did not contain the new parameters - the force push should have fixed this

theresa-cordero:
Note that this update of MOM6 will change answers if some defaults are not explicitly set. Otherwise we have confirmed that these updates do not change answers for NWA12 and NEP10 experiments.

PR #245, opened 2025-09-25
turn on automodule gha

yichengt900:
Try turn on automodule gha again......


uwagura:
@yichengt900 I'm not sure why the containerized run isn't reproducing on c5, as I believe @theresa-cordero was able to run this successfully the other day, but I can look into this issue from here

theresa-cordero:
Yes, I ran the container successfully on C5 yesterday.

uwagura:
Addressed by #247

Issue #251, opened 2025-12-02
Need a long term solution for example datasets


andrew-c-ross:
In a number of places, the documentation and scripts point to datasets stored at ftp.gfdl.noaa.gov:/pub/Yi-cheng.Teng. However, these datasets are currently unavailable because data is periodically swept from the public FTP server. We also don't want to keep bugging @yichengt900 to put the data back up. So, we should find a permanent place to put these datasets and update the documentation and scripts accordingly. We may also need @yichengt900 to point to where the authoritative versions of these datasets currently are within GFDL's storage network.

uwagura:
I think I may have come up with a more permanent solution for now: I created a sharable globus link that has the tar files on it. @andrew-c-ross , can you test this out and let me know what you think of it as a replacement for ftp: `wget https://g-8b4b27.fce040.8443.data.globus.org/1d_ci_datasets.tar.gz`

uwagura:
hmm did running `wget` on that link not work? The data should be public to all, so I'm not even sure why it asked for a log in on the web platform, but I certainly wasn't asked for a log in when I tried to wget the data.

andrew-c-ross:
For me, trying it on my personal laptop on guest wifi, wget downloaded the html for the login page 😖 

edit: It downloads a .tar.gz file with the right name, but when I look inside it's actually the html for the login page. Did you check the download that you got?

PR #255, opened 2026-01-16
Change FTP address of 1d datasets and point users to email for other datasets

uwagura:
Partially addresses #251 , along with #254 . This PR updates the ftp address for the 1d data sets from Yi-Cheng's `/pub` folder to a more permanent location on GFDL's data portal. Going forward, we should point folks to that location if they want our 1d datasets

Since we rarely get requests for the other datasets and they are much larger than the 1d dataset, I updated `exps/README` to tell users to reach out to us via email if they would like the larger datasets. It may be easier to serve these individually via globus or some other solution. If anyone thinks it's important to make those datasets publicly available, I can look into a more permanent solution for those as well.

### DIFF
--- a/builds/README.md
+++ b/builds/README.md
@@ -57,6 +57,6 @@ brew install wget
 
 ## Test run: 1-D MOM6-COBALT
 - To test your `MOM6SIS2`, first navigate to the `exps` folder: `cd ../exps`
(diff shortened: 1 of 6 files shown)


andrew-c-ross (approved):
Looks good to me. I tried downloading with the new URL from outside of the NOAA network and confirmed it is accessible.

PR #272, opened 2026-05-11
Add -j flag to linux-build.bash so that builds are done in parallel

uwagura:
This PR adds the `-j` flag to `$makeflags`, so that all builds are now done in parallel. In my tests on gaea, this cut the model build time by more than half: 
Without the `-j` flag: 
```
135759991                                   MOM6SIS2_gaea_build   gfdl_med        128  COMPLETED   00:17:20 
```
With the `-j` flag:
```
135761241                          MOM6SIS2_gaea_parallel_build   gfdl_med        128  COMPLETED   00:06:58
```

### DIFF
--- a/builds/linux-build.bash
+++ b/builds/linux-build.bash
@@ -85,7 +85,7 @@ build_libyaml() {
     pushd $srcdir/libyaml
     $srcdir/libyaml/bootstrap
     $srcdir/libyaml/configure --prefix="$target_dir" --disable-shared
-    make && make install
+    make -j && make install
     if [ $? -ne 0 ]; then
         echo "Could not build the libyaml library!"
         exit 1
--- a/builds/linux-build.bash
+++ b/builds/linux-build.bash
@@ -54,7 +54,7 @@ source $MODULESHOME/init/bash
 source $rootdir/$machine_name/$platform.env
 
 # Default Makeflags
-makeflags="NETCDF=4"
+makeflags="-j NETCDF=4"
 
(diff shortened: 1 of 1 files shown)


uwagura:
The effect is more subtle on the CI than on gaea. For this build with the -j flag, the build times were
```
build_mom6sis2_debug: 3 min 16s
build_mom6solo_repro: 7 min 30s
```
Whereas for the last build: 
```
build_mom6sis2_debug: 3 min 54s
build_mom6solo_repro: 9 min 30s
```
will look into this before merging @theresa-cordero

uwagura:
Latest run with parallel build of libyaml is even faster: 
```
build_mom6sis2_debug: 2 min 39s
build_mom6solo_repro: 6 min 34s
```
I think the reason why I'm not seeing the > 50% speed up I saw with the full repro build on gaea is 1.) that second build may not have included libyaml ( which does not seem to be rebuilt if the `.libs` directory is present ) and 2.) building cobalt serially in repro mode is a bigger bottleneck during that build process than serialization is to the `mom6sis2_debug` build or the `mom6solo_repro` build.  

Likewise, the time differences in the `mom6sis2` and `mom6solo` builds can likely be explained by the fact that one is built in debug mode, and one is built in repro.

theresa-cordero:
Is it safe to do a squash merge here or should we keep two commits?

theresa-cordero (approved):

uwagura:
I think  a squash merge should be fine

Issue #274, opened 2026-05-15
Slow atmoshere loop in NEP


uwagura:
I recently noticed that the atmosphere loop in the NEP10.COBALT configuration takes up almost the entire runtime of that configuration on c5. In particular, calculating fluxes up and down from the atmosphere seems to be extremely expensive processes: 
```
                                      hits          tmin          tmax          tavg          tstd  tfrac grain pemin pemax
Total runtime                            1   1735.926161   1735.937246   1735.931680      0.004344  1.000     0     0  2035
ATM                                     96   1467.428245   1467.501515   1467.492764      0.010582  0.845     0     0  2035
 ATM: atmos loop                        48   1385.819424   1455.948536   1444.268689     21.565762  0.832     0     0  2035
  A-L: atmos_tracer_driver_gathe         0      0.000000      0.000000      0.000000      0.000000  0.000     0     0  2035
  A-L: sfc_boundary_layer               48    172.626038    255.936415    236.115004     15.607382  0.136     0     0  2035
  A-L: update_atmos_model_dynami         0      0.000000      0.000000      0.000000      0.000000  0.000     0     0  2035
  A-L: serial radiation                 48      0.000003      0.000018      0.000006      0.000002  0.000     0     0  2035
  A-L: update_atmos_model_down           0      0.000000      0.000000      0.000000      0.000000  0.000     0     0  2035
  A-L: flux_down_from_atmos             48    261.119321   1013.277889    777.646164     69.253613  0.448     0     0  2035
  A-L: update_land_model_fast            0      0.000000      0.000000      0.000000      0.000000  0.000     0     0  2035
  A-L: update_ice_model_fast            48      0.009072      0.047030      0.013555      0.006043  0.000     0     0  2035
  A-L: flux_up_to_atmos                 48      0.003164    711.985274    199.232288    103.127466  0.115     0     0  2035
  A-L: update_atmos_model_up             0      0.000000      0.000000      0.000000      0.000000  0.000     0     0  2035
  A-L: update_atmos_model_state         48      0.000005      0.000194      0.000010      0.000005  0.000     0     0  2035
 ATM: update_land_model_slow             0      0.000000      0.000000      0.000000      0.000000  0.000     0     0  2035
Ocean                                   51    245.864845    246.260457    245.984180      0.074581  0.142     1     0  2035
```
[...]

PR #276, opened 2026-05-29
Add Script to automate bisect testing

uwagura:
This PR adds two scripts, `bissect-test.sh` and `do-bissect.sh`, to automate the process of doing a git bissect whenever there is an answer change. `bisect-test.sh` is the actual script called by `git bisect`, `do-bissect.sh` is just a wrapper that lets you run the bisect within a batch job. 

I vibe-coded these scripts a while back, but I wanted to wait until #274 was resolved and #272 was merged in so that building and running the models occurred in a reasonable enough amount of time to make an automated script useful. With those two issues resolved, I was able to use these scripts to correctly identify the answer change caused by commit `1e9ac3988b2a1c15c03841436ca8b95c4917de88` in between the last two `dev/gfd` -> `dev/cefi` merges in less than an hour. I have not tested the bisect process with any answer changes caused by the NWA, but now that this configuration runs in about 450 seconds for the 48 hour test run, I doubt it would take too long either.

This PR also adds a -n flag to `ci_build_driver.sh` to preserve the existing build directory in between bisect runs. @theresa-cordero , I know you were working on changes to this script to accomplish something similar, let me know if you want to merge in those changes first or maybe even add them to this PR. 

I added these scripts to the `builds` directory, but that choice was somewhat arbitrary. If folks think there is a better location for these scripts, let me know and we can move them there.

### DIFF
--- a/builds/bisect-test.sh
+++ b/builds/bisect-test.sh
@@ -1,0 +1,270 @@
+#!/bin/bash
+# Git bisect test script for CEFI-regional-MOM6 experiments
+#
+# Intended for tracking answer-changing commits in MOM6. For each commit
+# git bisect checks out, this script:
+#   1. Optionally merges a CEFI base commit on top of the checked-out commit
+#      so that CEFI-specific prerequisites are always present
+#   2. Builds MOM6SIS2 using the docker/linux-intel container settings
+#   3. Runs the 48hr test for the specified experiment
+#   4. Compares ocean.stats against a reference copy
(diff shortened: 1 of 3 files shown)


uwagura:
Since these scripts are highly specific to testing merges, I decided to move them to out internal git repo rather than adding them to the public repo. They are now available here: https://gitlab.gfdl.noaa.gov/cefi/cefi-bissect-script

PR #278, opened 2026-06-04
Correctly mask area for use in hypoxic area calculation

andrew-c-ross:
There was a bug in the hypoxic area diagnostic: instead of the cell area being masked to only hypoxic conditions and summed, oxygen was masked to hypoxic conditions, multiplied by area, and then summed. Fortunately, because the oxygen was masked to where it was < 2, this had a fairly minor impact on the diagnostic result. Nevertheless, this PR fixes the bug and correctly calculates the area where oxygen is < 2 mg/l.


uwagura (approved):
I had some difficulty running this on analysis, mostly due to the size of the data and perhaps because of some dmget issues, but I was eventually able to run it by extracting a few days worth of daily data and running on that rather than the whole thing. We may want to explore ways of improving some of the diagnostics that use large, daily datasets in future, but that can be for a later PR. This one is good to go for now.

PR #280, opened 2026-06-10
Update FMS and the coupler to the 2026.01.01 tag

uwagura:
As titled. This PR updates FMS and the coupler to their latest versions to keep up with their development


uwagura:
The 1d case is crashing in the container, but running as expected on gaea, with this update. Will investigate

uwagura:
The 1d crash was due to an issue in how `gfortran 11`, which is used to compile the model in our `base` container, handles `target` dummy variables that were introduced in `FMS 2026.01.01`

The short of it is that when a subroutine has a variable declared target: 
```
subroutine put_side1_to_xgrid(d, grid_id, x, xmap, remap_method, complete)
  use, intrinsic :: iso_c_binding, only: c_ptr, c_null_ptr, c_loc
  real(r8_kind), target, contiguous, intent(in)    :: d(:,:) !< data to send
```
And that subroutine is called with a variable that is not declared target:
```
    call fms_xgrid_put_to_xgrid (Atm%t_bot , 'ATM', ex_t_atm , xmap_sfc, remap_method=remap_method, complete=.false.)
    call fms_xgrid_put_to_xgrid (Atm%z_bot , 'ATM', ex_z_atm , xmap_sfc, remap_method=remap_method, complete=.false.)
    call fms_xgrid_put_to_xgrid (Atm%p_bot , 'ATM', ex_p_atm , xmap_sfc, remap_method=remap_method, complete=.false.)
    call fms_xgrid_put_to_xgrid (Atm%u_bot , 'ATM', ex_u_atm , xmap_sfc, remap_method=remap_method, complete=.false.)
    call fms_xgrid_put_to_xgrid (Atm%v_bot , 'ATM', ex_v_atm , xmap_sfc, remap_method=remap_method, complete=.false.)
    call fms_xgrid_put_to_xgrid (Atm%p_surf, 'ATM', ex_p_surf, xmap_sfc, remap_method=remap_method, complete=.false.)
    call fms_xgrid_put_to_xgrid (Atm%slp,    'ATM', ex_slp,    xmap_sfc, remap_method=remap_method, complete=.false.)
    call fms_xgrid_put_to_xgrid (Atm%gust,   'ATM', ex_gust,   xmap_sfc, remap_method=remap_method, complete=.true.)

```
You technically get undefined behavior, as per section 15.5.2.5.11 of the  2024 FORTRAN standard: 
```
If the dummy argument has the TARGET attribute and the effective argument 
does not have the TARGET attribute or is an array section with a vector subscript, 
any pointers associated with the dummy argument become undefined when 
execution of the procedure completes.
```
It seems as though gfortran 11 handled this issue by creating a copy of the non-target argument ( in this case, the `Atm%` variables from above), and then freeing that copy after the end of the call to the subroutine. However, it appears as though these temporary copies were later referenced in the when `complete=.true` in the`xgrid` code from above leading to a use-after-free error in gfortran 11 that manifested itself in the 1d test case as junk values being read into the surface temperatures:
```
Bad temperatures (dimension 1):            1           1           1           1           1           1           1           1           1           1           1           1           1           1           1           1

FATAL: compute_qs: saturation vapor pressure table overflow, nbad=     16
```
( Disclaimer: This explanation was largely put together with the help of an AI) 

Newer versions of gfortran, and ifort, do not seem to have this issue, so we can get around this issue ( and preserve answers) by just updating to a newer gfortran. PR #281 will have to be merged in before this one to accomplish this.

theresa-cordero (approved):

PR #281, opened 2026-06-12
Update container to Ubuntu 2024.04 with gfortran 13, update tides reference answers

### LINKED ISSUE #258: Update Ubuntu in CI container after figuring out why answers change
Our CI container currently uses Ubuntu 22.04, which will reach it's end of life in April 2027. PR #257 showed that answers change in the 2d tides test case when we run the container using Ubuntu 24.04, which prevented us from updating the OS in that PR. We need to figure out why answers change for this test case ( and don't change for the other test cases ) so that we can update the container OS before it goes EOL.

uwagura:
Fixes #258 . #280 revealed that the latest version of FMS and the coupler cannot run the 1D case with gfortran 11 due to a discrepancy in how that version of gfortran handles dummy variables declared `target` introduced in FMS 2026.01.01 and how new versions of gfortran/ifort handle dummy variables declared `target`. This means it's probably time that we bite the bullet and update our container to Ubuntu 2024.04 with gfortran 13.3. 

I unfortunately was not able to figure out why the answers changed in the tidesonly test case, and rather than waste more Claude tokens chasing this down, I decided we're better off just accepting the new answers. If folks think we should spend more time on this before accepting the new answers, let me know. I've moved the original answers to a `g11` directory inside of the `NWA12.tidesonly/ref` directory so that we can reinvestigate this in future if need be. Similarly, the original Ubuntu 2022 container is still available under the original `base` tag in the packages within this repo. 

This PR also modifies the oxygen budget closure test to write the output from the `ncap2` command to file, as for some reason the test was failing within this container when we didn't write the output to a file.


.github/workflows/ci-test-mom6-cobalt.yaml
```diff
@@ -72,15 +72,15 @@ jobs:
           ncwa -a xh,yh o2_col_budget.nc -O o2_col_budget.nc
           # Check residual for surface layer
           ncks -d zl,0 o2_col_budget.nc o2_sfc_budget.nc
-          resid=$( ncap2 -s 'result = o2h_tendency -
+          resid=$( ncap2 -O -s 'result = o2h_tendency -
```
theresa-cordero, line 75: This change seems unrelated to updating the container, if you think it is an important change could you put it in a different PR?

theresa-cordero: Or just add a comment to this PR to say why it is related!

uwagura: @theresa-cordero , for some reason, the oxygen budget closure test failed in the new container because we were not writing the output from that `ncap2` to a file. I don't quite know why it failed this time and not when @andrew-c-ross originally wrote it ( maybe some sort of nco versioning issue? ), but this was the simplest fix I could find.  

Without this change, the budget closure test will not run in the new container, so I think it's ok to lump it in with this PR, but I can update the PR description to include this detail for documentation purposes.

EDIT: Failed run here: https://github.com/NOAA-GFDL/CEFI-regional-MOM6/actions/runs/27431844485/job/81084366913

theresa-cordero (approved):

PR #282, opened 2026-06-17
Update SIS to latest dev/gfdl

uwagura:
As titled


uwagura:
For some reason, the 1d test case cannot push changes to the parameter doc files to my branch, which is why the previous CI failed. I went ahead and pushed the changes manually just to get the CI past this run, but going forward this is something we will have to address.

theresa-cordero (approved):

PR #283, opened 2026-06-17
Auto-update submodules to latest dev/cefi (auto/update-submodules-20260617)

yichengt900:
This PR automatically updates the specified submodules to the latest commit on their `dev/cefi` branch.

**Updated Submodules:**

- src/ocean_BGC (updated to commit 901a726f1f4f2c4021ef7d76aa1521b275bcd94d)

---
Auto-generated by GitHub Actions.


uwagura:
The NWA test case failed because a new variable was added to the restart file: 
```
compare ice_model.res.nc
DIFFER : Failed to find variable h_pond_ice in file "/gpfs/f5/gfdl_med/proj-shared/github/ci_data/reference/main/NWA12.COBALT/20260407/ice_model.res.nc".
```
It looks like this was added by NOAA-GFDL/SIS2#226. All other restart files are otherwise identical. 

I think we can just go ahead and update the reference `ice_model.res` file to include this new variable and this PR should be good to go @theresa-cordero 

Not sure what happened with the NEP test, but in a separate test it seem to reproduce answers as well.

theresa-cordero (approved):

PR #288, opened 2026-07-27
Fix a handful of minor issues in boundary condition creation workflow

andrew-c-ross:
As titled, this fixes:
1. A typo in mom6_obc_workflow.sh
2. An error loading miniforge because of using $MODULESHOME/init/sh within a bash script
3. A bug where ./ouputs is always created even if the config sets the output dir to somewhere else (and if so, that other directory needs to be created)


uwagura (approved):
Looks good to me. I was able to run `mom6_obc_workflow.sh` on ppan with no issues, so I think this PR is good to go.

PR #289, opened 2026-07-31
Get all source code from CEFI-regional-MOM6 and share compile experiment across XMLs

### LINKED ISSUE #287: Duplicate repository clones in XMLs
The XMLs clone the CEFI-regional-MOM6.git repository, which includes submodules for all of the source code components, but then each component also has a source block that clones the component repository again. As a result we end up with two copies of FMS, coupler, ice_param, land_null, and atmos_null, and it is easy to confuse which copy actually gets compiled. I would like to clean this up first before thinking about switching the coupler to point to our custom version with the wind/current coefficient. 

The problem is that FRE demands a source block with a codeBase block pointing to a git repository for each model component. 

My proposed solution to this is to 
1. Move the CEFI-regional-MOM6 clone to the first component, FMS.
```xml
<component name="fms" paths="FMS">
  <source versionControl="git" root="https://github.com/NOAA-GFDL/">
    <codeBase version="$(CEFI_GIT_TAG)">CEFI-regional-MOM6.git</codeBase>
    <csh>
    <!-- next step will go here -->
    </csh>
  </source>
  <compile>
    <cppDefs>-Duse_libMPI -Duse_netCDF $(F2003_FLAGS) -DMAXFIELDMETHODS_=600</cppDefs>
  </compile>
</component>
```
2. Use the csh block above to create a nearly empty local git repository that meets the bare minimum expectations of FRE (repository with a named branch):
```xml
<csh><![CDATA[
  mv CEFI-regional-MOM6 CRM6
  git init -b main null.git
  (cd null.git && git commit --allow-empty -m empty)
]]></csh>
```
3. For all of the components after FMS, the paths will point to the CEFI-regional-MOM6 src (renamed CRM6 above), and the source block will clone the local null.git and then delete it right away. For example, this would be the coupler:
```xml
<component name="coupler" paths="CRM6/src/coupler/full CRM6/src/coupler/shared" requires="fms land_null atmos_null sis2 mom6_symmetric">
  <description domainName="FMS Coupler" communityName="coupler"/>
  <source versionControl="git" root=".">
    <codeBase version="main"> null.git </codeBase>
    <csh><![CDATA[ rm -rf null]]></csh>
  </source>
  <compile>
    <cppDefs><![CDATA[$(F2003_FLAGS) -DUSE_FMS2_IO -D_USE_LEGACY_LAND_ -Duse_AM3_physics -D_FILE_VERSION="'"`git-version-string $<`"'"]]></cppDefs>
  </compile>
</component>
```

In addition to eliminating the duplication, this will allow consistent versions of all of the code components to be checked out by just setting the `CEFI_GIT_TAG` parameter (no more `FMS_GIT_TAG`, `CPL_GIT_TAG`, `NULL_TAG`, etc). 

The downside is that this is an ugly hack, but I couldn't find any cleaner way around FRE's requirement to have a git repo for each component. 

What do you think @theresa-cordero and @uwagura ?

andrew-c-ross:
This closes #287 by using the approach I proposed there and @yichengt900 improved. The FMS component clones the CEFI-regional-MOM6 repository and checks out the specified branch/tag, and all of the other components do not clone any additional code. 

While cleaning up the XMLs, I took the opportunity to move all of the components to a shared XML. In components.xml, each component is placed inside its own freInclude. In the actual experiment XMLs, the component XML is defined as an entity
```xml
<!DOCTYPE experimentSuite [
  <!ENTITY include_components SYSTEM "../components.xml">
]>
```
and then inserted using `&include_components;`. Then the specific components can be included using <xi:include>. This reduces the compile experiment to:
```xml
<experiment name="$(COMPILE_EXP)">
  <description>
    Make the executable for ocean-ice experiments.
    Each component is defined as a freInclude ultimately sourced from ../components.xml.
    To check out a specific release of the CEFI code base, set CEFI_GIT_TAG.
    To change individual components, delete the FMS include,
    copy the FMS comopnent from ../compoments.xml,
    and customize the component root or csh.
  </description>
  <xi:include xpointer="xpointer(//freInclude[@name='FMS']/component)"/>
  <xi:include xpointer="xpointer(//freInclude[@name='MOM6_SYMMETRIC']/component)"/>
  <xi:include xpointer="xpointer(//freInclude[@name='SIS2']/component)"/>
  <xi:include xpointer="xpointer(//freInclude[@name='LAND_NULL']/component)"/>
  <xi:include xpointer="xpointer(//freInclude[@name='ATMOS_NULL']/component)"/>
  <xi:include xpointer="xpointer(//freInclude[@name='COUPLER']/component)"/>
</experiment>
```

I would have liked to point the <xi:include> directly to the components.xml, but I couldn't find a way to do this without hitting an error caused by an additional and unexpected attribute getting injected into the component when including using href. 

One thing to check is the NEP XML, which includes the Icepack component. I tried to preserve how this was defined previously, but does it actually get used?

### DIFF
--- a/xmls/NEP10/CEFI_NEP_cobalt.xml
+++ b/xmls/NEP10/CEFI_NEP_cobalt.xml
@@ -1,4 +1,7 @@
 <?xml version="1.0"?>
+<!DOCTYPE experimentSuite [
+  <!ENTITY include_components SYSTEM "../components.xml">
+]>
 <!-- FRE Usage documentation: http://www.gfdl.noaa.gov/fms/fre -->
 <!--
 Quick start guide
@@ -34,30 +37,9 @@ frecheck -v -r restart -p ncrc6.intel23 -t repro  -x CEFI_NEP_cobalt.xml CEFI_NE
 
 
 
-<experimentSuite rtsVersion="4" xmlns:xi="http://www.w3.org/2001/XInclude">
-  <!-- git tag for general component source code-->
-  <property name="RELEASE" value="2024.02"/>
-
-  <!-- git tag for FMS source code -->
-  <property name="FMS_GIT_TAG" value="2025.03"/>
-
-  <!-- git tag for coupler source code -->
-  <property name="CPL_GIT_TAG" value="2025.03"/>
-
-  <!-- git tag for null (atmos_null, land_null, etc) source code -->
-  <property name="NULL_TAG" value="2025.02"/>
-
-  <!-- git tag for ocean_shared source code -->
-  <property name="OCEAN_BGC_GIT_TAG" value="dev/cefi"/>
-
-  <!-- Updated MOM6 source code commit to checkout -->
-  <property name="MOM6_GIT_TAG" value="main"/>
-
-  <!-- Updated Icepack commit to check out -->
-  <property name="ICEPACK_GIT_TAG" value="Icepack1.3.3"/>
-
+<experimentSuite rtsVersion="4" xmlns:xi="http://www.w3.org/2003/XInclude">
   <!-- customize as needed -->
-  <property name="FRE_STEM" value="fre/cefi/NEP/2025_06"/>
+  <property name="FRE_STEM" value="fre/cefi/NEP/2026_07"/>
 
   <!-- Please make sure to change your group, such as b, f, g, m, o... -->
   <!-- NCRC_GROUP will only have impact for your root dir and work dir on gaea c5/c6-->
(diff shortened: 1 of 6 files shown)


theresa-cordero:
> One thing to check is the NEP XML, which includes the Icepack component. I tried to preserve how this was defined previously, but does it actually get used?

I don't believe the NEP uses Icepack, it is probably left over from when we were testing the sea-ice ridging (which comes from Icepack). Since it was listed as an include, but the path wasn't pointed to I think you change is fine, but I will test it. 

On the other hand, the ARC11k should be compiled with Icepack, but we could add that once we have apublished version with solutions we are tracking. Aside from taking more time to compile and being one more code base to track, I don't think there is a downside to including Icepack in all the builds.

xmls/NEP10/CEFI_NEP_cobalt.xml
uwagura: We need to change the name of the compile experiments in the sample commands to match the new names

xmls/NWA12/CEFI_NWA12_physics.xml
```diff
@@ -108,79 +96,33 @@ frecheck -r layout -p ncrc6.intel23 -x CEFI_NWA12_physics.xml CEFI_NWA12_physics
...
-  <property name="PROG_MAIN" value="coupler/coupler_main.o"/>
-  <property name="LIBS_ROOT" value="MOM6_SIS2_GENERIC_4P_compile_symm"/>
-  <property name="FMS_LIB_DIR" value="./$(LIBS_ROOT)/$(platform)-$(target)/exec"/>
+  <!-- COMPILE -->
 
-  <property name="MY_LIBS" value="$(EXEC_ROOT)/$(FMS_LIB_DIR)"/>
+  <!-- git tag for CEFI-regional-MOM6 source code -->
+  <property name="CEFI_GIT_TAG" value="main"/>
+  <!-- Name of the compile experiment: -->
+  <property name="COMPILE_EXP" value="FMS2_MOM6_SIS2_COBALT_compile_symm"/>
```
uwagura: Maybe we should just name the physics only xml `FMS2_MOM6_SIS2_compile_symm` so that if uses are running both physics only and BGC experiments they can keep track of separate executables with and without COBALT

xmls/NEP10/CEFI_NEP_cobalt.xml
```diff
@@ -114,88 +96,34 @@ frecheck -v -r restart -p ncrc6.intel23 -t repro  -x CEFI_NEP_cobalt.xml CEFI_NE
...
-  <property name="MY_LIBS" value="$(EXEC_ROOT)/$(FMS_LIB_DIR)"/>
-
-  <experiment name="MOM6_SIS2_GENERIC_4P_compile_symm">
+  <experiment name="$(COMPILE_EXP)">
     <description>
       Make the executable for ocean-ice experiments.
+      Each component is defined as a freInclude ultimately sourced from ../components.xml.
+      To check out a specific release of the CEFI code base, set CEFI_GIT_TAG.
+      To change individual components, delete the FMS include,
+      copy the FMS comopnent from ../compoments.xml,
```
uwagura:
```suggestion
      copy the FMS component from ../components.xml,
```

xmls/NWA12/CEFI_NWA12_cobalt.xml
```diff
@@ -112,79 +100,33 @@ frecheck -r layout -p ncrc6.intel23 -x CEFI_NWA12_cobalt.xml CEFI_NWA12_COBALT_V
...
 
-  <experiment name="MOM6_SIS2_GENERIC_4P_compile_symm">
+  <experiment name="$(COMPILE_EXP)">
     <description>
       Make the executable for ocean-ice experiments.
-      Uses FMS2 (for unknown reason FMS1 does not work with intel23 on C6)
+      Each component is defined as a freInclude ultimately sourced from ../components.xml.
+      To check out a specific release of the CEFI code base, set CEFI_GIT_TAG.
+      To change individual components, delete the FMS include,
+      copy the FMS comopnent from ../compoments.xml,
```
uwagura:
```suggestion
      copy the FMS component from ../components.xml,
```

xmls/NWA12/CEFI_NWA12_cobalt_fms2_yaml.xml
```diff
@@ -112,79 +100,57 @@ frecheck -r layout -p ncrc6.intel23 -x CEFI_NWA12_cobalt_fms2_yaml.xml CEFI_NWA1
...
 
-  <experiment name="MOM6_SIS2_GENERIC_4P_compile_symm_yaml">
+  <experiment name="$(COMPILE_EXP)">
     <description>
       Make the executable for ocean-ice experiments.
-      Uses FMS2 (for unknown reason FMS1 does not work with intel23 on C6)
+      Each component is defined as a freInclude ultimately sourced from ../components.xml.
+      To check out a specific release of the CEFI code base, set CEFI_GIT_TAG.
+      To change individual components, delete the FMS include,
+      copy the FMS comopnent from ../compoments.xml,
```
uwagura:
```suggestion
      copy the FMS component from ../components.xml,
```

xmls/NWA12/CEFI_NWA12_physics.xml
```diff
@@ -108,79 +96,33 @@ frecheck -r layout -p ncrc6.intel23 -x CEFI_NWA12_physics.xml CEFI_NWA12_physics
...
 
-  <experiment name="MOM6_SIS2_GENERIC_4P_compile_symm">
+  <experiment name="$(COMPILE_EXP)">
     <description>
       Make the executable for ocean-ice experiments.
-      Uses FMS2 (for unknown reason FMS1 does not work with intel23 on C6)
+      Each component is defined as a freInclude ultimately sourced from ../components.xml.
+      To check out a specific release of the CEFI code base, set CEFI_GIT_TAG.
+      To change individual components, delete the FMS include,
+      copy the FMS comopnent from ../compoments.xml,
```
uwagura:
```suggestion
      copy the FMS component from ../components.xml,
```

xmls/NWA12/CEFI_NWA12_physics.xml
andrew-c-ross: :+1:  I named the physics-only compile experiment FMS2_MOM6_SIS2_compile_symm. I also renamed the MOM6 include to MOM6_BGC, and added a physics-only component included using MOM6_ONLY. Inside the include, both name the component 'mom6' so that the other components requiring MOM6 will recognize whichever is used.

uwagura (approved):
I was able to compile all four executables on c6 with ifort. The new symlinks seem to be working as expected. I think this PR is good to go.

theresa-cordero:
I am happy with these changes and look forward to not having duplicate src directories!
I think there are some outstanding questions about Icepack, but since the NWA and NEP do not use Icepack code we can modify the compile instructions to compile Icepack whenever we have a model that uses it.

theresa-cordero (approved):

PR #290, opened 2026-08-03
Point coupler to experimental wind/current branch

andrew-c-ross:
This changes the coupler submodule to point to the NOAA-CEFI-Regional-Ocean-Modeling fork instead of NOAA-GFDL, and also uses my branch with the experimental wind/current modification. 

The new branch adds the namelist option `wind_current_coefficient` to allow scaling of the ocean surface current the coupler subtracts from the wind.
```xml
<namelist name="surface_flux_nml">
  wind_current_coef = 0.5
</namelist>
```

I didn't change any of the XMLs in this PR, and when not set the coefficient defaults to 1.0 which should reproduce old results. This will make it easier to switch to using the coefficient though.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -44,5 +44,5 @@
 	branch = 2025.03
 [submodule "src/coupler"]
 	path = src/coupler
-	url = https://github.com/NOAA-GFDL/FMScoupler.git
-	branch = 2025.03
+	url = https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/FMScoupler.git
+	branch = feature/wind_current_coef


uwagura:
I think this change is a good idea. @theresa-cordero and I have been tracking sub module commits by simply committing a particular tag of each sub-module to a new branch and then merging these changes in once they past testing. This PR is now making me realize that we haven't been updating the `.gitmodules` file in the process, which is now out of date for most components. It doesn't appear as though this causes any issues, but it may be worth updating.

Perhaps we should have one PR that updates that file for all submodules, and then another PR which points the version of the coupler in `src/coupler` to the CEFI branch?  Let me know what you guys think.

andrew-c-ross:
:+1: Updating all the submodules first, as well as starting to tag stable versions more frequently, would be especially helpful now that we've set up the XMLs to rely on a single checkout of CEFI-regional-MOM6 for all of the source code components

theresa-cordero:
Yes, I think we should update the `.gitmodules` first then merge in this PR.

I support pointing to this version of the coupler in the CEFI src directory

PR #291, opened 2026-08-20
Change NWA set_visc_answer_date to 20260703 to restore answers

uwagura:
As titled. NOAA-GFDL/MOM6#1143 changed the implementation of `find_L_open_convex` in `MOM_set_viscosity.F90` to use a reproducible cuberoot function for `set_visc_answer_date >= 20260704` . This PR changes the default answer date to be the day before this. Our NWA12.COBALT `MOM_input` used the old default answer date, so running that experiment after this PR was merged in would select the new `cuberoot` over the old power operator and answers would change.

My initial plan for this PR was to switch out our current `MOM_input`, which is based on the `MOM_parameter_doc.all` file, for the `MOM_parameter_doc.short` file, but it looks like there are some parameters that are not logged in the short file such as `TS_CONFIG`, so using the short file causes the model to crash. Rather than manually add all necessary parameters to the `MOM_parameter_doc.short`, I'm just gonna modify our current `MOM_input` while noting this exposes us to this issue in future if a default answer date is changed from `99990101` to a current date.


theresa-cordero:
I believe this experiment always starts from a restart so the initialization settings such as `TS_CONFIG` are not read or saved in the `MOM_param_doc.short`. If you switch to a new run in the input.nml they should be written.

What error do you get when `TS_CONFIG` is missing? For a restart run I am surprised the default value isn't okay for the model to run. 

This could be a MOM6 bug, since you should be able to recreate an experiment using the `.short`

andrew-c-ross:
@theresa-cordero in my recent XMLs I've had to add `#override TS_CONFIG="file"` even though the runs are all starting from spun up restart files. This is the only line I had to add, so I never bothered trying to find what had changed to make it required, but there may have been a bug introduced into MOM6.

uwagura:
It looks like when starting a simulation from RESTARTS, TS_CONFIG is not logged if you start a run from restarts, and the model will not crash if it is not present in an input file: 
```fortran
    ! Initialize temperature and salinity (T and S).
    if ( use_temperature ) then
      call get_param(PF, mdl, "TS_CONFIG", config, &
             "A string that determines how the initial temperatures "//&
             "and salinities are specified for a new run: \n"//&
             " \t file - read velocities from the file specified \n"//&
             " \t\t by (TS_FILE). \n"//&
             " \t fit - find the temperatures that are consistent with \n"//&
             " \t\t the layer densities and salinity S_REF. \n"//&
             " \t TS_profile - use temperature and salinity profiles \n"//&
             " \t\t (read from TS_FILE) to set layer densities. \n"//&
             " \t benchmark - use the benchmark test case T & S. \n"//&
             " \t linear - linear in logical layer space. \n"//&
             " \t DOME2D - 2D DOME initialization. \n"//&
             " \t ISOMIP - ISOMIP initialization. \n"//&
             " \t adjustment2d - 2d lock exchange T/S ICs. \n"//&
             " \t sloshing - sloshing mode T/S ICs. \n"//&
             " \t seamount - no motion test with seamount ICs. \n"//&
             " \t dumbbell - sloshing channel ICs. \n"//&
             " \t rossby_front - a mixed layer front in thermal wind balance.\n"//&
             " \t SCM_CVMix_tests - used in the SCM CVMix tests.\n"//&
             " \t USER - call a user modified routine.", &
             fail_if_missing=new_sim, do_not_log=just_read)
!            " \t baroclinic_zone - an analytic baroclinic zone. \n"//&

```
As a result, TS_CONFIG wasn't in the short param doc, and the model did not fail in spite of this fact, so it didn't read anything and filled the `config` variable with garbage. This gave me this difficult to understand error message when I used the short param doc: 
```
FATAL from PE   950: MOM_initialize_state: Unrecognized Temp & salt configuration 0190101 result in the use of older, less accurate expressions If true, cell-by-cell reconstructions are checked for consistency and if non-monotonicity or an inconsistency is If true, use partial cell
```
This is probably something worth fixing upstream in MOM6. In the meantime, I think I can probably add TS_CONFIG to our short doc and get something working

~~EDIT: I also don think any of the initialization code should even be executed if we are starting from restarts, but that is another issue~~, never mind, it looks like nothing is actually written to variables stored within the model since `just_read = .true`

uwagura:
@theresa-cordero The reason the run crashes is because there is no default value for `TS_CONFIG`. I set a value for TS_CONFIG in the MOM_input so that the run would work, but we may need to add a default value to the code to prevent this from happening in future. 

As a side note, the shorter MOM_input seems to speed up NWA initialization a bit, though we are still spending a significant fraction of the NWA runtime initialization the model: 
old c6 times
```
                                      hits          tmin          tmax          tavg          tstd  tfrac grain pemin pemax
Total runtime                            1    431.665074    431.665272    431.665148      0.000051  1.000     0     0  1599
Initialization                           1    252.888145    252.890817    252.889854      0.000665  0.586     0     0  1599
  Init: atmos_model_init                 1      0.201993      0.202018      0.202006      0.000003  0.000     0     0  1599
  Init: land_model_init                  1      0.114575      0.241493      0.200264      0.032673  0.000     0     0  1599
  Init: ice_model_init                   1      6.042940      6.046522      6.043804      0.000384  0.014     0     0  1599
  Init: ocean_model_init                 1    242.377398    242.380064    242.378357      0.000515  0.561     0     0  1599
  Init: flux_exchange_init               1      0.747252      0.747322      0.747272      0.000005  0.002     0     0  1599
```
new c6 times
```
                                      hits          tmin          tmax          tavg          tstd  tfrac grain pemin pemax
Total runtime                            1    408.681806    408.682059    408.681915      0.000071  1.000     0     0  1599
Initialization                           1    232.033485    232.037761    232.036761      0.000872  0.568     0     0  1599
  Init: atmos_model_init                 1      0.176196      0.176213      0.176201      0.000003  0.000     0     0  1599
  Init: land_model_init                  1      0.120388      0.241054      0.199427      0.031394  0.000     0     0  1599
  Init: ice_model_init                   1      5.736519      5.741914      5.738446      0.000761  0.014     0     0  1599
  Init: ocean_model_init                 1    222.550940    222.553588    222.552249      0.000592  0.545     0     0  1599
  Init: flux_exchange_init               1      0.636697      0.636725      0.636717      0.000004  0.002     0     0  1599

```

uwagura:
Now that NOAA-GFDL/MOM6#1170 is merged in, this PR no longer requires setting `TS_CONFIG` in `MOM_input` and has been updated to not set this parameter.

exps/NWA12.COBALT/INPUT/MOM_input
```diff
@@ -344,196 +143,69 @@ OBC_SEGMENT_003_VELOCITY_NUDGING_TIMESCALES = 3.0, 360.0 !   [days]
...
-                                ! If true, use the pseudo salt tracer, typically run as a diagnostic.
-USE_BOUNDARY_IMPULSE_TRACER = False !   [Boolean] default = False
-                                ! If true, use the boundary impulse tracer.
-USE_DYED_OBC_TRACER = False     !   [Boolean] default = False
-                                ! If true, use the dyed_obc_tracer tracer package.
-USE_NW2_TRACERS = False         !   [Boolean] default = False
-                                ! If true, use the NeverWorld2 tracers.
 
 ! === module register_MOM_generic_tracer ===
-GENERIC_TRACER_IC_FILE = "bgc_woa_esper_ics_1993_2023-04.nc" ! default = ""
```
theresa-cordero, line 514: Do we need to keep this line if `GENERIC_TRACER_IC_FILE=""` is removed from `MOM_override`?

exps/NWA12.COBALT/INPUT/MOM_override
theresa-cordero, line 1: I think all the overrides in this file are unnecessary with the current `MOM_input`.  

The only exception is `GENERIC_TRACER_IC_FILE`, which probably shouldn't be needed since this is a restart run.

uwagura: I think you're right. `GENERIC_TRACER_IC_FILE` defaults to a blank string anyways, so even if this wasn't a run from restarts we wouldn't need that line in our override.

andrew-c-ross:
What would you think about keeping the MOM_input based on the parameter doc .all instead of .short? I've always preferred to use the .all as the basis for the input to avoid any surprises and prevent changing defaults from having undesired effects.

uwagura:
@andrew-c-ross We actually wanted to switch to the .short because the old default `SET_VISC_ANSWER_DATE = 99990101` in the .all file caused an unexpected answer change by selecting a newer version of the cuberoot instead of the version we were previously using. We hoped that using the short would prevent this from happening in the future, assuming that future default changes are non-answer changing.

andrew-c-ross:
Would it be better to set the various answer dates to something near the present date instead of the current mix of old and far-future dates? In hindsight for a published/stable configuration that makes more sense to me.

uwagura:
I think @theresa-cordero suggested something along these lines earlier, and I think it would address both of our concerns. We need a date before `20260704` to get around the `SET_VISC_ANSWER_DATE` issue. Maybe we could set all `99990101` answer dates to `20260703`, and then leave answer dates that are earlier than this?

theresa-cordero:
Yeah, I suggested using the doc all but with answer dates near the present day. We could also use `20260703` or even a date closer to the publication date.

uwagura:
> Yeah, I suggested using the doc all but with answer dates near the present day. We could also use `20260703` or even a date closer to the publication date.

This change has now been implemented, and this PR is ready for review

theresa-cordero (approved):

PR #296, opened 2026-09-03
refactor/to python module

andrew-c-ross:
- Moves a lot of code that is common across different scripts into one module.
- Preserves most of the code and functionality, except for a few changes to account for updated dependencies or other minor changes.
- Uses uv for installing the new module and its dependencies; more modern and drops dependence on conda
- Includes HCtFlood as an optional dependency so we don't need to hack around that by adding it to sys.path

This is a stacked PR, so the next PRs in the stack will improve the formatting of the code and fix some issues.


tools/initial/write_glorys_initial.py
```diff
@@ -56,20 +46,20 @@ def write_initial(config):
...
     revert = glorys.interp(depth=ztarget, kwargs={'fill_value': 'extrapolate'}).ffill('zl', limit=None)
 
-    # Flood temperature and salinity over land. 
+    # Flood temperature and salinity over land.
     flooded = xarray.merge((
         flood.flood_kara(revert[v], zdim='zl') for v in [temp_var, sal_var, u_var, v_var]
     ))
 
     # flood zos separately to avoid the extra z=0 added by flood_kara.
     flooded[ssh_var] = flood.flood_kara(revert[ssh_var]).isel(z=0).drop('z')
```
theresa-cordero: While we are doing this refactor, could we switch to `drop_vars` to stop this warning `FutureWarning: dropping variables using `drop` is deprecated; use drop_vars.`?

PR #301, opened 2026-09-16
Update drivers and build script to default to c6, baremetal ifx testing

uwagura:
This PR updates the `driver.sh` scripts in the `NEP10.COBALT` and `NWA12.COBALT` configurations to default to paths on c6 rather than c5, and updates the `linux-build.bash` script to compile for `ifx` on `c6` by default. Given that c6 has been cefi's production machine for quite some time now, we will lose access to c5 in about a year, and ifx is the new default compiler, this seems like the appropriate time to move things over. 

Although there is a copy of our regression testing container available on c6, I've elected not to use it for testing. We've had issues in the past with the container failing on gaea ( for example, this was the reason for #243 ), and the current container only has ifx version 2023.2.0 anyways, so it will require an update before we can use it for testing. Maintaining a container seems like an extra layer of complexity we don't really need, though let me know if anyone feels otherwise. Containers may be worth exploring again if we ever do testing on the cloud.

There is a workflow file on our private gitlab mirror that I've confirmed executes with these changes


theresa-cordero:
I agree with moving to C6 for testing. 

If we are planning to stop supporting the container, should we archive or put a warning label on `ci_build_driver.sh`? I do use that script for regression testing, but I am fine with just using `linux-build.bash`.

uwagura:
> I agree with moving to C6 for testing.
> 
> If we are planning to stop supporting the container, should we archive or put a warning label on `ci_build_driver.sh`? I do use that script for regression testing, but I am fine with just using `linux-build.bash`.

I've updated `ci_build_driver.sh` to just call `linux-build.bash`, so you can keep using it for regression testing and any script we may have that rely on it can still work ( for example, my bissect script ).

theresa-cordero (approved):

